PR #122, opened 2021-10-19
fix-release-gitignore

jazzyfresh:
ignore top level pkg directory created by release automation, but include internal/pkg


jrasell (approved):
LGTM!

PR #168, opened 2021-11-09
Render to Files

### LINKED ISSUE #109: Add ability to render to files
When using Pack in CI/CD, it would be great to be able to render to a file.

Something like `nomad-pack render <PACK> --to=./path/to/dir` maybe?

angrycub:
- Implement --render-to-file
- Add non-interactive ErrorWithContext
- Format Error Context Elements
- Add auto-approve flag
- Update to --to-dir; adopt --auto-approve flag for overwrite
- Implemented Input in glintUI
- Missed the callsite renaming confirmOverwrite

Closes #109

### DIFF
--- a/cli/render.go
+++ b/cli/render.go
@@ -1,11 +1,19 @@
 package cli
 
 import (
+	stdErrors "errors"
+	"fmt"
+	"io/fs"
+	"io/ioutil"
+	"os"
+	"path"
+	"strings"
+
 	v1 "github.com/hashicorp/nomad-openapi/v1"
 	"github.com/hashicorp/nomad-pack/flag"
 	"github.com/hashicorp/nomad-pack/internal/pkg/cache"
 	"github.com/hashicorp/nomad-pack/internal/pkg/errors"
-	"github.com/mitchellh/go-glint"
+	"github.com/hashicorp/nomad-pack/terminal"
 	"github.com/posener/complete"
 )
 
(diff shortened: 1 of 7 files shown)


tgross:
@angrycub is there any way to get the render to work with shell redirection, rather than requiring command line flags? I'm not sure what the heck glint is doing to break this but it seems like a basic Unix thing we shouldn't be breaking.

angrycub:
A pack can render multiple output files, which wouldn't stream into a single file in a meaningful way.

tgross:
Oh, that's what I was missing then!

cli/render.go
```diff
@@ -15,9 +24,134 @@ import (
...
+	// Check to see if the file already exists and validate against the value
+	// of overwrite.
+
+	_, err := os.Stat(path)
+
+	if err == nil && !overwrite {
+		return fmt.Errorf("destination file exists and overwrite is unset")
+	}
+
+	err = ioutil.WriteFile(path, []byte(content), 0644)
```
DerekStrickland, line 140: `ioutil` is deprecated and goes away as of go 1.17 I believe. Should we update this now to use `os` functions to be future proof?

cli/render.go
```diff
@@ -15,9 +24,134 @@ import (
...
+		return nil
+	}
+	info, err := os.Stat(path)
+
+	if err != nil {
+		if stdErrors.Is(err, fs.ErrNotExist) {
+			return nil
+		}
+
+		return fmt.Errorf("unexpected error testing --to-dir path: %w", err)
```
DerekStrickland, line 106:
```suggestion
		return fmt.Errorf("unexpected error validating --to-dir path: %w", err)
```

cli/render.go
```diff
@@ -15,9 +24,134 @@ import (
 type RenderCommand struct {
 	*baseCommand
 	packConfig *cache.PackConfig
-	// renderOutputTemplate is a boolean flag to control whether the outpu
+	// renderOutputTemplate is a boolean flag to control whether the output
 	// template is rendered.
 	renderOutputTemplate bool
+	// renderToDir is the path to write rendered job files to in addition to
+	// standard output.
+	renderToDir string
```
DerekStrickland, line 32: I'm not seeing where we add this to the ErrorContext. Am I just missing it in this view?  Would be useful to add if we are not yet.

cli/render.go
```diff
@@ -64,16 +202,17 @@ func (c *RenderCommand) Run(args []string) int {
...
+	// Iterate the rendered files and add these to the list of renders to
+	// output. This allows errors to surface and end things without emitting
+	// partial output and then erroring out.
 
-	// Iterate the rendered files and add these to the Glint document.
-	// TODO(jrasell): trim at least the templates directory name from the name
-	//  as it doesn't provide much benefit.
 	for name, renderedFile := range renderOutput.DependentRenders() {
-		addRenderToDoc(d, name, renderedFile)
+		renders = append(renders, Render{Name: formatRenderName(name), Content: renderedFile})
```
DerekStrickland, line 212: How do you feel about adding these Names to the ErrorContext? Same for ParentRenders.

terminal/glint.go
```diff
@@ -34,15 +40,48 @@ func (ui *glintUI) Close() error {
 }
 
 func (ui *glintUI) Input(input *Input) (string, error) {
-	return "", ErrNonInteractive
+	var buf bytes.Buffer
```
DerekStrickland, line 43: Nice addition!

terminal/glint.go
```diff
@@ -34,15 +40,48 @@ func (ui *glintUI) Close() error {
...
+	}
 }
 
 // Interactive implements UI
 func (ui *glintUI) Interactive() bool {
-	// TODO(mitchellh): We can make this interactive later but Glint itself
-	// doesn't support input yet. We can pause the document, do some input,
-	// then resume potentially.
-	return false
+	return isatty.IsTerminal(os.Stdin.Fd())
```
DerekStrickland, line 84: If glint still doesn't support interactive, maybe you could submit an upstream PR. Worth it?

terminal/noninteractive.go
```diff
@@ -205,7 +205,19 @@ func (ui *nonInteractiveUI) Error(msg string) {
 // ErrorWithContext satisfies the ErrorWithContext function on the UI
 // interface.
 func (ui *nonInteractiveUI) ErrorWithContext(err error, sub string, ctx ...string) {
-	ErrorWithContext(err, sub, ctx...)
+	ui.Error(strings.Title(sub))
```
DerekStrickland, line 208: Nice

DerekStrickland (commented):
Left some feedback. Nothing blocking. Happy to approve when you promote from WIP to ready for review.

DerekStrickland (approved):
:shipit:

PR #181, opened 2022-01-05
cli: run template canonicalization before planning to fix diffs.

### LINKED ISSUE #14: cli: plan should run canonicalization on job specification before planning
When planning a deployed job without changes, the returned diff includes a removal for the custom Nomad Pack meta parameters. This is because job canonicalization adds additional data to a job specification including Nomad Pack metadata and has the potential to set Consul/Vault tokens and namespaces.

The plan command should therefore be updated to include additional flags that can modify the job specification, alongside calling the job canonicalization function, before submitting the plan.

jrasell:
closes #14


angrycub (approved):
LGTM

PR #183, opened 2022-01-06
Improve Error Reporting on API Client errors

### LINKED ISSUE #164: some errors returned are obtuse
the newly added jenkins pack contains an error and when i try to run it the error returned by nomad-pack is less than useful.  I had to render the pack and manually and run the job, to get presented with some reasonable feedback from the system.

to reproduce.

> packs ❯ pwd 
> nomad-pack-community-registry/packs
packs ❯ nomad-pack run jenkins -f jenkins/vars.nomad
! Failed To Parse Job Specification

	Error:   400 Bad Request
	Type:    client.GenericOpenAPIError
	Context:
	         - Template Name: jenkins/templates/jenkins.nomad.tpl
	         - Registry Name: dev
	         - Pack Name: jenkins
	         - Pack Ref: dev
	         - Deployment Name: jenkins
> packs ❯ nomad-pack render jenkins -f jenkins/vars.nomad

safe the output to a file (see #163 for my bug report about not being able to pass directly to file) and run it using the nomad cli

> packs ❯ nomad run jenkins.nomad              
> Error getting job struct: Failed to parse using HCL 2. Use the HCL 1 parser with `nomad run -hcl1`, or address the following issues:
> jenkins.nomad:61,7-13: Unsupported argument; An argument named "config" is not expected here. Did you mean to define a block of type "config"?

### LINKED ISSUE #180: nomad-pack plan returns 500 error
I created a custom pack. I was working from local files instead of from a repository. When tried to `plan` the pack, it returned the following message:
```
$ nomad-pack plan ./packs/test_app/ --var-file=/Users/SunSparc/test/nomad-pack-test-vars.hcl
! Failed To Perform Plan

  Error:   500 Internal Server Error
  Type:    client.GenericOpenAPIError
  Context:
           - Registry Name: dev
           - Pack Name: test_app
           - Pack Ref: dev
           - Deployment Name: test_app
           - Template Name: test_app/templates/test-app.nomad.tpl
           - Job Name: nomad-test-app
```

When I `plan` the pack without the `var-file` there is no error:
```
$ nomad-pack plan ./packs/test_app/
+ Job: "nomad-test-app"
+ Task Group: "group1" (1 create)
  + Task: "task1" (forces create)  + Task: "task2" (forces create)  + Task: "task3" (forces create)

» Scheduler dry-run:
Plan succeeded
```

I tried the `plan` by specifying some variables as parameters and those worked also:
```
nomad-pack plan ./packs/test_app/ --var app_version="1.2.3" --var data_version="2.3.4" --var var1_test="test value" --var ver2_test=3.14
```

As I was writing this issue and going through different iterations, I narrowed the problem down to the `region` variable. I had given the `region` variable a value that was apparently not correct. I corrected the value and the `plan` succeeded.

Perhaps this issue falls along the same lines as #164.

angrycub:
Added an unwrapping method to get the Body() information as the error text for `GenericOpenAPIError`.

Closes #180
Closes #164

### DIFF
--- a/fixtures/bad_pack/README.md
+++ b/fixtures/bad_pack/README.md
@@ -1,0 +1,114 @@
+# Simple Service
+
+This pack is a used to deploy a Docker image to as a service job to Nomad.
+
+This is ideal for configuring and deploying a simple web application to Nomad.
+
+## Customizing the Docker Image
+
+The docker image deployed can be replaced with a variable. In the example
+below, we will deploy and run `httpd:latest`.
+
+```
+nomad-pack run simple_service --var image="httpd:latest"
+```
+
+## Customizing Ports
+
+The ports that are exposed via Docker can be customized as well.
+
+In this case, we'll write the port values to a file called `./overrides.hcl`:
+
+```
+{
+  name = "http"
+  port = 8000
+},
+{
+  name = "https"
+  port = 8001
+}
+```
+
+Then pass the file into the run command:
+
+```
+nomad-pack run simple_service -f ./overrides.hcl`"
+```
+
+## Customizing Resources
+
+The application resource limits can be customized:
+
+```
+resources = {
+  cpu = 500
+  memory = 501
+}
+```
+
+## Customizing Environment Variables
+
+Environment variables can be added:
+
+```
+env_vars = [
+  {
+    key = "foo"
+    value = 1
+  }
+]
+```
+
+## Consul Service and Load Balancer Integration
+
+Optionally, this pack can configure a Consul service.
+
+If the `register_consul_service` is unset or set to true, the Consul service will be registered.
+
+Several load balancers in the [The Nomad Pack Community Registry](../README.md) are configured to connect to
+this service with ease.
+
+The [NginX](../nginx/README.md) and [HAProxy](../haproxy/README.md) packs can be configured to balance over the
+Consul service deployed by this pack. Just ensure that the "consul_service_name" variable provided to those
+packs matches this consul_service_name.
+
+The [Fabio](../fabio/README.md) and [Traefik](../traefik/README.md) packs are configured to search for Consul
+services with the specific tags.
+
+To tag this Consul service to work with Fabio, add `"urlprefix-<PATH>"`
+to the consul_tags. For instance, to route at the root path, you would add `"urlprefix-/"`. To route at the path `"/api/v1"`, you would add '"urlprefix-/api/v1".
+
+To tag this Consul service to work with Traefik, add "traefik.enable=true" to the consul_tags, also add "traefik.http.routers.http.rule=Path(\`<PATH>\`)". To route at the root path, you would add "traefik.http.routers.http.rule=Path(\`/\`)". To route at the path "/api/v1", you would add "traefik.http.routers.http.rule=Path(\`/api/v1\`)".
+
+```
+register_consul_service = true
+
+consul_tags = [
+  "urlprefix-/",
+  "traefik.enable=true",
+  "traefik.http.routers.http.rule=Path(`/`)",
+]
+```
+
+## Customizing Consul and Upstream Services
+
+Consul configuration can be tweaked and (upstream services)[https://www.nomadproject.io/docs/job-specification/upstreams]
+can be added as well.
+
+```
+register_consul_service = true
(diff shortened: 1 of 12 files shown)


cli/cli_test.go
```diff
@@ -110,13 +111,20 @@ func TestJobPlan(t *testing.T) {
...
+}
+
 // Confirm that another pack with the same job names but a different deployment name fails
 func TestJobPlanConflictingDeployment(t *testing.T) {
 	testInit(t)
-
-	defer func() {
-		clearJob(t, &cache.PackConfig{Name: testPack})
-	}()
+	defer clearJob(t, &cache.PackConfig{Name: testPack})
```
DerekStrickland, line 127: love this!  I did this somewhere else, and missed doing it here I guess.

cli/cli_test.go
```diff
@@ -125,15 +133,13 @@ func TestJobPlanConflictingDeployment(t *testing.T) {
 	exitCode = runCmd().Run([]string{testPack, testRefFlag})
 	require.Equal(t, 1, exitCode)
 
-	reset()
+	// reset is run in clearJob
```
DerekStrickland, line 136: Not blocking, but I'd be ok just deleting these comment lines.

cli/cli_test.go
```diff
@@ -365,8 +370,25 @@ func testInit(t *testing.T) {
 	nomadAddr = os.Getenv("NOMAD_ADDR")
 	_ = os.Setenv("NOMAD_ADDR", "http://127.0.0.1:4646")
 
+	_, err := os.Stat(path.Join(cache.DefaultCachePath(), cache.DefaultRegistryName, "simple_service@latest"))
```
DerekStrickland, line 373: Just out of curiosity, why do we need this new block?

cli/cli_test.go
```diff
@@ -397,11 +419,16 @@ func clearJob(t *testing.T, cfg *cache.PackConfig) {
 }
 
 func nomadExec(t *testing.T, args ...string) {
+	var outb, errb bytes.Buffer
+
 	nomadPath, err := exec.LookPath("nomad")
 	require.NoError(t, err)
 	nomadCmd := exec.Command(nomadPath, args...)
+	nomadCmd.Stdout = &outb
```
DerekStrickland, line 427: Very nice.

cli/helpers.go
```diff
@@ -9,6 +9,7 @@ import (
 	v1 "github.com/hashicorp/nomad-openapi/v1"
 	"github.com/hashicorp/nomad-pack/internal/pkg/cache"
 	"github.com/hashicorp/nomad-pack/internal/pkg/errors"
+	intHelper "github.com/hashicorp/nomad-pack/internal/pkg/helper"
```
DerekStrickland, line 12: haha, I had to double-take this because I read it as "integer helper" internally and then was surprised by its use.

fixtures/bad_pack/README.md
```diff
@@ -0,0 +1,114 @@
+# Simple Service
```
DerekStrickland, line 1: +1000 on adding this fixture.

DerekStrickland (approved):
LGTM. I had some questions, but no blockers.

cli/cli_test.go
angrycub: I thought the comment might help prevent folks from putting a reset at the end of these tests and re-breaking things.  The other possibility could be moving the reset up to a defer on the cases that don't require `clearJob` so that the end action is always up at the top of the test func.  What do you think?

cli/cli_test.go
angrycub: When I ran the tests with an empty packs folder, it would not download the latest registry without adding this case.  This caused all of the tests that used `simple_service@latest` to fail. I could have worked around it by cloning default first myself, but that seems to be a surprising prerequisite to running the tests.

cli/cli_test.go
DerekStrickland: Yeah, I like the idea of the end action in a defer at the top of the function, and then engineers can decide what to call there.

cli/cli_test.go
DerekStrickland: 😮  

That seems like something to file an issue for, yeah?

PR #184, opened 2022-01-13
Fixed render's overwrite prompt

angrycub:
- moved approve to its own flagset so it can be used w/o the other ones
- added an "a" to the overwrite options so that all overwrites can be set at once interactively

### DIFF
--- a/cli/render.go
+++ b/cli/render.go
@@ -30,6 +30,8 @@ type RenderCommand struct {
 	// renderToDir is the path to write rendered job files to in addition to
 	// standard output.
 	renderToDir string
+	// overwriteAll is set to true when someone specifies "a" to the y/n/a
+	overwriteAll bool
 }
 
 type Render struct {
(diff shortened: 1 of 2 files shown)


cli/render.go
```diff
@@ -75,21 +68,36 @@ func (r Render) toFile(c *RenderCommand, ec *errors.UIErrorContext) error {
...
 		if err != nil {
 			return false, err
 		}
 		overwrite = strings.ToLower(overwrite)
-		if overwrite == "y" || overwrite == "n" {
-			return overwrite == "y", nil
+		switch overwrite {
+		case "a":
+			c.overwriteAll = true
+			fallthrough
```
lgfa29, line 94: Is the idea to `return true, nil` like in the `y` case? If so I think it would be better to use the explicit `return` to avoid a potential bug if a new `case` gets added or the order changes.

lgfa29: Side note not related to this PR: is the `for` loop used to guard against invalid options and ask the question again? If that's the case it would be nice to have `default` with some kind of `invalid option, please try again`.

cli/render.go
```diff
@@ -127,14 +135,20 @@ func maybeCreateDestinationDir(path string) error {
 	return nil
 }
 
-func writeFile(path string, content string, overwrite bool) error {
+func writeFile(c *RenderCommand, path string, content string) error {
 	// Check to see if the file already exists and validate against the value
 	// of overwrite.
+	var overwrite bool
```
lgfa29, line 141: We don't use this value outside the `if err == nil {` scope right? So we can probably just walrus (`:=`) it there.

lgfa29 (approved):
Minor comments, but no blockers 🙂

PR #185, opened 2022-01-14
Alias standard library errors funcs into internal errors package

angrycub:
Aliasing these functions allows developers to not have to include one or the other package under an import alias.

Added a `newError` alias to the `New` function to make creating errors inside of the package a little more explicit.

Refactored code to use alias methods.

### DIFF
--- a/internal/pkg/errors/wrapped_context_test.go
+++ b/internal/pkg/errors/wrapped_context_test.go
@@ -4,7 +4,6 @@ import (
 	"testing"
 
 	"github.com/hashicorp/hcl/v2"
-
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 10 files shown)


internal/pkg/errors/wrapped_context_test.go
```diff
@@ -1,10 +1,10 @@
 package errors
 
 import (
-	stdErrors "errors"
-	"github.com/hashicorp/hcl/v2"
 	"testing"
 
+	"github.com/hashicorp/hcl/v2"
+
```
jrasell, line 7:
```suggestion
```

jrasell (approved):
LGTM, thanks @angrycub!

PR #190, opened 2022-01-27
Remove dev binary from $GOBIN before copying over

angrycub:
For some reason, the current behavior leaves me with a trashed binary that dies. Removing
it before a copy that was going to squash it anyway seems okay.


DerekStrickland (approved):
:shipit:

PR #191, opened 2022-01-28
Handle HEREDOC delimiter immediately before EOF

### LINKED ISSUE #189: Heredoc unsupported inside included vars.hcl files?
I'm writing my own version of the traefik pack (needs a lot of modification for internal use) and this:

```
variable "toml_config" {
	type = string
	description = "traefik.toml config"
	default     = <<EOF
[entryPoints]
  [entryPoints.http]
  address = ":8080"
  [entryPoints.traefik]
  address = ":1936"
[api]
  dashboard = true
  insecure = true
[providers.consulCatalog]
  prefix           = "traefik"
  exposedByDefault = false
[providers.consulCatalog.endpoint]
  address = "{{ env "attr.unique.network.ip-address" }}:8500"
  scheme  = "http"
EOF
}
```

Produces exactly that block, no problems.

However, when doing `nomad-pack render -f vars.hcl .` where vars.hcl contains
```
toml_config = <<EOF
[example]
  toml-config
EOF
```
The error is

```
! Failed To Process Pack

	Error:   No closing marker was found for the string.
	Type:    *errors.errorString
	Context:
	         - HCL Range: test.hcl:28,4-4
	         - Registry Name: dev
	         - Pack Name: traefik
	         - Pack Ref: dev
```
I've gotten through everything else but this problem without too much trouble, but this one really has me stuck...

I can use a string literal here (and I am, for now) but that's really unmaintainable long-term.

angrycub:
Adding a linefeed to the end of the file ensures that terminal heredoc end delimiters are processed properly. This is a bit of a hack, but is the lowest touch fix.

Fixes #189

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## Unreleased
 
 BUG FIXES:
-
+* template: Handle HEREDOC delimiter immediately before EOF [GH-191](https://github.com/hashicorp/nomad-pack/pull/191)
 
 IMPROVEMENTS:
 
--- a/fixtures/variables-with-heredoc/README.md
+++ b/fixtures/variables-with-heredoc/README.md
@@ -1,0 +1,19 @@
+# Test Pack for heredocs in varfiles
+
+Reported Error:
+
+```
+nomad-pack render toml_pack -f vars.hcl
+! Failed To Process Pack
+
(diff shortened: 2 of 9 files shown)


jrasell (approved):
LGTM!

note: needs a changelog entry.

PR #192, opened 2022-01-31
Move flag to internal/pkg

angrycub:
To continue to make the nomad-pack source tree more waypoint shaped, this PR moves the flag package from the root of the project to `internal/pkg`


DerekStrickland (approved):
:shipit:

PR #196, opened 2022-02-01
Fix CLI and flag tests.

angrycub:
This PR is to:
* **Fix flag tests** - The flags in the test were malformed and so wouldn't add, causing the test to fail. Fixed and added a table-driven test with some more cases.

* **Fix cli tests**
    - After the refactor in #193, the relative path to the fixture folder changed. Corrected that in appropriate tests.
    - Standardized all test cleanup calls to be immediately run in a defer after `testInit` is called. Removed extra `func()` wrapper from `defer`s that don't need it
    - Updated the funcs that call the `nomad` cli binary to handle cases where the:
        - CLI is expected to return an error.
        - CLI is expected to not return an error.
        - caller doesn't care—specifically `clearJob`.
    - Added a custom error type to allow tests to collect `stderr` and `stdout` from runs of the CLI tool to facilitate test writing/debugging.

### DIFF
--- a/internal/cli/cli_test.go
+++ b/internal/cli/cli_test.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"os"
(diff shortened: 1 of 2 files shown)


DerekStrickland (approved):
:shipit:

PR #197, opened 2022-02-02
Upgrade Sprig to v3

angrycub:
The website documents their head, so we were missing some funcs (`get` for example).


DerekStrickland (approved):
:shipit:

PR #201, opened 2022-02-03
Support setting variables via environment

angrycub:
- Make single variable fixture from heredoc pack
- Enable `NOMAD_PACK_VAR_xxx` Environment Variables


internal/pkg/variable/cli.go
```diff
@@ -9,6 +9,76 @@ import (
...
+		packVarName[0] = splitName[0]
+		packVarName[1] = splitName[1]
+	default:
+		// We cannot handle a splitName where the variable includes more than
+		// one separator.
+		return hcl.Diagnostics{
+			{
+				Severity: hcl.DiagError,
+				Summary:  "Invalid NOMAD_PACK_var option",
+				Detail:   fmt.Sprintf("The given environment variable NOMAD_PACK_var_%s=%s is not correctly specified. The variable name must not have more than forward-slash separators.", name, rawVal),
```
DerekStrickland, line 33:
```suggestion
				Detail:   fmt.Sprintf("The given environment variable NOMAD_PACK_var_%s=%s is not correctly specified. The variable name must not have more than one period separator.", name, rawVal),
```
I "think" you meant this? Also, remind me, will we ever?

internal/pkg/variable/cli.go
```diff
@@ -9,6 +9,76 @@ import (
...
+	}
+
+	val, diags := expr.Value(nil)
+	if diags.HasErrors() {
+		return diags
+	}
+
+	// If our stored type isn't cty.NilType then attempt to covert the override
+	// variable, so we know they are compatible.
+	if existing.Type != cty.NilType {
```
DerekStrickland, line 62: if it is nil, should we just exit and not append anything? Or do we need to append nils for some downstream processing?

DerekStrickland (approved):
Asked some questions, but didn't see any blockers.

internal/pkg/variable/cli.go
angrycub: Ugh... copypasta.  Taking this moment to use the constant in the message in case we decide at a future date to change it.

lgfa29 (approved):

PR #209, opened 2022-02-23
Use c.UI for version command

angrycub:
**Description**
Switch from glint implementation to c.UI implementation for version.  Will allow non-interactive UIs to work properly with the version command.


DerekStrickland (approved):
Non-blocking historical context.

At one point we had talked about having a "CI" envar similar to how `npm` works to trigger alternate behavior and then have a factory method that would give you back a CI-friendly UI versus the Glint UI. I don't know if we ever chatted about that, but I suspect we have because of logging. Not sure if it's still part of the roadmap, but I thought I'd share.

PR #212, opened 2022-02-24
Add operations flagset to status command for --name

angrycub:
**Description**
The `TestStatusFails` test was supposed to be validating that providing the `--name` flag absent a pack name was an error; however, the test currently only looks for a non-successful error code.  In this case, the `--name` flag was missing from the status command, but since that error also returned exitcode 1 the test was passing anyway.

I've updated the command, and there is a big test improvement PR coming that will enable us to check the returned errors from command calls.

**Reminders**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## UNRELEASED
 
 BUG FIXES:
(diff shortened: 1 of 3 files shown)


jrasell (approved):
LGTM.

PR #215, opened 2022-03-02
chore: Cleanup lint issues

angrycub:
**Description**
- Fix lint issues
- Export flag helpers to fix deadcode


internal/cli/stop.go
```diff
@@ -103,7 +106,8 @@ func (c *StopCommand) Run(args []string) int {
 
 			// get job struct from template
 			// TODO: Should we add an hcl1 flag?
```
DerekStrickland, line 108: I would say yes, we should add it.

DerekStrickland (approved):
Looks great. Couple of questions but just for myself.  Nothing blocking.

PR #217, opened 2022-03-03
Use Pack metadata `Name` in error context once known.

### LINKED ISSUE #214: I found that getting 'pack name' seems to be a problem
When I execute ` nomad pack render` 'pack name' gets' metadata 'name' field defined in HCl file

However, I specify ` nomad pack render when the file is executed- f ./ my-variables. HCl ` what it gets is my directory name

```
[root@d es]# nomad-pack info .
Pack Name          wordpress
```

```
[root@d es]# cat templates/wordpress.nomad.tpl 
job [[ template "job_name" . ]] {
  [[ template "region" . ]]
  datacenters = [ [[ range $idx, $dc := .wordpress.datacenters ]][[if $idx]],[[end]][[ $dc | quote ]][[ end ]] ]
  type = "service"

```

```
[root@d es]# nomad-pack render .  
wordpress/wordpress.nomad:

job "wordpress" {
  namespace = "elasticsearch"
  
  datacenters = [ "dc1" ]
  type = "service"
  update {
    max_parallel     = 1
... ...
```

```
[root@d es]# nomad-pack render .  -f my-variables.hcl 
! Failed To Process Pack

	Error:   There is no variable named "mariadb_group_health_check". An override file can only override a variable that was already declared in a primary configuration file.
	Type:    *errors.errorString
	Context: 
	         - HCL Range: my-variables.hcl:32,1-36,2
	         - Registry Name: dev
	         - Pack Name: es
	         - Pack Ref: dev
```

angrycub:
**Description**
- Add test fixtures for `nomad-pack #214` - #214
- Less likely sentinel values for pack from folder
- Add pack path to error context for pack from file
- Store `loadedPack` in pack manager
- Make `Add` an upsert in error context
- Retrieve packname from manager; upsert into err context
- Add changelog

**Reminders**

Closes #214

### DIFF
--- a/internal/pkg/cache/cache.go
+++ b/internal/pkg/cache/cache.go
@@ -15,8 +15,8 @@ const (
 	DefaultRegistryName   = "default"
 	DefaultRegistrySource = "github.com/hashicorp/nomad-pack-community-registry"
 	DefaultRef            = "latest"
-	DevRegistryName       = "dev"
-	DevRef                = "dev"
+	DevRegistryName       = "<<local folder>>"
+	DevRef                = "<<none>>"
 )
 
 // NewCache instantiates a new cache instance with the specified config. If no
--- a/internal/pkg/errors/error_context.go
+++ b/internal/pkg/errors/error_context.go
@@ -13,8 +13,20 @@ type ErrorContext struct {
 // NewErrorContext creates an empty ErrorContext.
 func NewErrorContext() *ErrorContext { return &ErrorContext{} }
 
-// Add formats and appends the passed prefix and value onto the error contexts.
+// Add formats and upserts the passed prefix and value onto the error contexts.
 func (ctx *ErrorContext) Add(prefix, val string) {
+	idx := -1
+	for i, c := range ctx.contexts {
+		if strings.HasPrefix(c, prefix) {
+			idx = i
+			break
+		}
+	}
+	if idx != -1 {
+		ctx.contexts[idx] = prefix + val
+		return
+	}
+
 	ctx.contexts = append(ctx.contexts, prefix+val)
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 BUG FIXES:
 * cli: Add missing --name flag for status command [[GH-212](https://github.com/hashicorp/nomad-pack/pull/212)]
+* cli: Use Pack metadata `Name` in error context once known [[GH-217](https://github.com/hashicorp/nomad-pack/pull/217)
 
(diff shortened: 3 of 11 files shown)


internal/pkg/errors/error_context.go
```diff
@@ -13,8 +13,20 @@ type ErrorContext struct {
...
 func NewErrorContext() *ErrorContext { return &ErrorContext{} }
 
-// Add formats and appends the passed prefix and value onto the error contexts.
+// Add formats and upserts the passed prefix and value onto the error contexts.
 func (ctx *ErrorContext) Add(prefix, val string) {
+	idx := -1
+	for i, c := range ctx.contexts {
+		if strings.HasPrefix(c, prefix) {
+			idx = i
+			break
```
lgfa29, line 22: Isn't it possible to `ctx.contexts[idx] = prefix + val` here directly and then `break`?

(and same for `ui_context.go`)

lgfa29 (approved):
Minor comment. I also pushed a commit to close the CHANGELOG entry link 👍

internal/pkg/cache/cache.go
```diff
@@ -15,8 +15,8 @@ const (
 	DefaultRegistryName   = "default"
 	DefaultRegistrySource = "github.com/hashicorp/nomad-pack-community-registry"
 	DefaultRef            = "latest"
-	DevRegistryName       = "dev"
-	DevRef                = "dev"
+	DevRegistryName       = "<<local folder>>"
```
DerekStrickland, line 18: I like this improvement a lot. This was super last minute addition and didn't get thought through as well as it should have.

internal/pkg/errors/error_context.go
```diff
@@ -13,8 +13,20 @@ type ErrorContext struct {
 // NewErrorContext creates an empty ErrorContext.
 func NewErrorContext() *ErrorContext { return &ErrorContext{} }
 
-// Add formats and appends the passed prefix and value onto the error contexts.
+// Add formats and upserts the passed prefix and value onto the error contexts.
 func (ctx *ErrorContext) Add(prefix, val string) {
+	idx := -1
```
DerekStrickland, line 18: +1

DerekStrickland (approved):
LGTM

internal/pkg/errors/error_context.go
angrycub: It is possible, however, in that case, the error context value is malformed.  The current concatenated value that is stored into context makes this issue very difficult to tease apart.  A future goal is to split these context values into a key/value pair, which would give us an opportunity to make this less _brittle_.

PR #227, opened 2022-03-28
Refactor cache_test for parallelism, local clones for speed

angrycub:
**Description**
This PR reorganizes pack tests so that they leverage a generated local repository for clones. This allows the tests to be run in parallel and reduces the demand for network on the runners.

### DIFF
--- a/internal/pkg/cache/cache_test.go
+++ b/internal/pkg/cache/cache_test.go
@@ -7,6 +7,7 @@ import (
 	"os/exec"
 	"path"
 	"strings"
+	"sync"
 	"testing"
 
 	"github.com/hashicorp/nomad-pack/internal/pkg/errors"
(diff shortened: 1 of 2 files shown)


internal/pkg/cache/cache_test.go
```diff
@@ -40,13 +49,116 @@ func testPackCount(t *testing.T, opts cacheOperationProvider) int {
...
+	git(handleGitError, "config", "user.email", "test@example.com")
+	git(handleGitError, "config", "user.name", "Github Action Test User")
+	git(handleGitError, "add", ".")
+	git(handleGitError, "commit", "-m", "Initial Commit")
+	res := git(handleGitError, "log", "-1", `--pretty=%H`)
+	*testRegistryRef1 = strings.TrimSpace(res.stdout)
+
+	git(handleGitError, "commit", "--allow-empty", "-m", "Second Commit")
+	res = git(handleGitError, "log", "-1", `--pretty=%H`)
+	*testRegistryRef2 = strings.TrimSpace(res.stdout)
```
lgfa29, line 122: Personal preference, but passing the error handler like this feels a bit odd to me. A more explicit error handling would look like

```go
res := git("init")
if res != nil {
    handleInitError(res)
}
res := git(, "config", "user.email", "test@example.com")
if res != nil {
    handleGitError(res)
}
// ...
```

But that's quite verbose. Another option would be to make `git` a stateful `struct` that you can just call and it would store errors as they happen:

```go
type git struct {
    res *GitCommandResult
}

func (g git) run(args ...string) GitCommandResult {
    if g.res != nil {
        // we've failed already, so noop
        return nil
    }
    git := exec.Command("git", args...)
    // ...
    err := git.Run()
    if err != nil {
        g.res = GitCommandResult { ... }
    }
}

func makeTestRegRepo() func() {
    // ...
    g := git{}
    g.run("init")
    if g.res != nil {
        // handle init error
    }

    g.run("config", "user.email", "test@example.com")
    g.run("config", "user.name", "Github Action Test User")
    g.run("add", ".")
    // ...
    if g.res != nil {
        formatPanic(res)
    }
    // ..
```

internal/pkg/cache/cache_test.go
```diff
@@ -403,3 +525,67 @@ func TestParsePackURL(t *testing.T) {
...
+// for any of the methods. This can be useful for cases that require a Logger as
+// a parameter.
+type NoopLogger struct{}
+
+func (_ NoopLogger) Debug(_ string)                                  {}
+func (_ NoopLogger) Error(_ string)                                  {}
+func (_ NoopLogger) ErrorWithContext(_ error, _ string, _ ...string) {}
+func (_ NoopLogger) Info(_ string)                                   {}
+func (_ NoopLogger) Trace(_ string)                                  {}
+func (_ NoopLogger) Warning(_ string)                                {}
```
lgfa29, line 591: Does this need to be defined here as well in addition to `internal/pkg/logging/logger.go`?

lgfa29 (approved):
Minor comments, but looks great 🙂

PR #230, opened 2022-03-28
Enable nomad-pack to read all client config vars from environment

angrycub:
**Description**
Currently, all of the nomad-pack options to configure its internal Nomad API client must be configured using the flags. This PR adds the corresponding environment variables and tests to validate proper precedence.

**Reminders**


DerekStrickland (approved):
:shipit:

PR #236, opened 2022-04-11
[cli] Add flags to change exit codes on plan command

### LINKED ISSUE #224: Nomad-pack plan has an exit code of 1
Successful nomad-pack plan returns an exit code of 1. This issue was discovered when using a CI/CD pipeline within GitHub actions, although a `continue-on-error: true` prevents the pipeline from failing it is not a viable long term solution.

```
❯ nomadpack plan . -f ./nomad_vars.hcl
Job: "mark"
Task Group: "app" (1 create/destroy update, 1 in-place update)
  Task: "server"

» Scheduler dry-run:
Plan succeeded

root in template_nomad_repo/packs/hello_world on  init 🏎💨  v1.1.5default took 2s
❯ echo $?
1
```

angrycub:
**Description**
Provide CLI flags to allow end users to specify the exit codes for the `plan` command.

Closes #224

**Reminders**

### DIFF
--- a/internal/cli/plan.go
+++ b/internal/cli/plan.go
@@ -11,11 +11,18 @@ import (
 
 type PlanCommand struct {
 	*baseCommand
-	packConfig *cache.PackConfig
-	jobConfig  *job.CLIConfig
+	packConfig        *cache.PackConfig
+	jobConfig         *job.CLIConfig
+	exitCodeNoChanges int
+	exitCodeChanges   int
+	exitCodeError     int
 }
 
 func (c *PlanCommand) Run(args []string) int {
+	c.exitCodeNoChanges = 0
+	c.exitCodeChanges = 1
+	c.exitCodeError = 255
+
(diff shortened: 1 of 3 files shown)


internal/cli/plan.go
```diff
@@ -177,6 +195,26 @@ func (c *PlanCommand) Flags() *flag.Sets {
...
+			Default: 1,
+			Usage:   `Override exit code returned when the plan shows changes.`,
+		})
+
+		f.IntVar(&flag.IntVar{
+			Name:    "exit-code-error",
+			Target:  &c.exitCodeError,
+			Default: 255,
+			Usage:   `Override exit code returned when there is an error.`,
+		})
```
lgfa29, line 217: One alternative that I thought would be to have `k=v` flag that can be defined multiple times instead of specific flags: 

```console
$ nomad-pack plan -exit-code "error=20" -exit-code "no-changes=2"
```

It seems like it would be easier to add new values, but I don't know if it would actually provide any real benefit.

lgfa29 (approved):
:shipit:

PR #238, opened 2022-04-12
docs: nested template requires `.` to pass variables

tgross:
Without including the `.` in the nested template caller, the variables
for the parent won't be rendered in the child templates.

Ran into this while working on https://github.com/hashicorp/nomad-pack-community-registry/pull/113 and was very confused :grinning:


angrycub (commented):
Good catch. There are definitely some places we need to try to catch up the docs, so PRs like this are very helpful!

angrycub (approved):

PR #240, opened 2022-04-14
docs: remove author and url fields from writing packs

tgross:
Originally `app.author` was intended to be the actual author of the
app, but many pack maintainers put their own names here. We have the
GitHub handles of the actual maintainers, so remove this field from
the guide.

Likewise, the `pack.url` field doesn't match the actual registry URL
if someone forks a registry. Remove that field as well.

ref conversation with @mikenomitch in https://github.com/hashicorp/nomad-pack-community-registry/pull/113#discussion_r848825515 and https://github.com/hashicorp/nomad-pack-community-registry/pull/113#discussion_r848827897


tgross:
As it turns out, `author` is required.

pkazmierczak:
Closing this since the issue has been resolved by #317 and #343.

PR #242, opened 2022-04-22
[chore] Onboard Nomad-Pack to common release toolkit

angrycub:
**Description**
Onboarding Nomad-Pack to the HashiCorp Common Release toolkit.

- Add version target to Makefile
- Add Docker build outputs
- Add CRT release configuration
- Add CRT build workflow


.github/workflows/build.yml
```diff
@@ -0,0 +1,268 @@
...
+
+      fail-fast: true
+
+    name: Go ${{ needs.get-go-version.outputs.go-version }} ${{ matrix.goos }} ${{ matrix.goarch }} build
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Setup go
+        uses: actions/setup-go@v2
```
lgfa29, line 90:
```suggestion
        uses: actions/setup-go@v3
```

I don't think it matters here, but I spent a long time until I realized that `v2` doesn't set `GOBIN` to `PATH`.

.release/ci.hcl
```diff
@@ -0,0 +1,275 @@
...
+  action "promote-production-packaging" {
+    organization = "hashicorp"
+    repository = "crt-workflows-common"
+    workflow = "promote-production-packaging"
+  }
+
+  notification {
+    on = "always"
+  }
+}
```
lgfa29, line 275: I don't know if we have an HCL linter in this repo, but we had to `hclfmt` this file in Nomad

Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+COPY dist/$TARGETOS/$TARGETARCH/nomad-pack /bin/
+
+# Create a non-root user to run the software.
+RUN addgroup $PRODUCT_NAME && \
+    adduser -S -G $PRODUCT_NAME $PRODUCT_NAME && \
+    mkdir -p /home/$PRODUCT_NAME/.cache && \
+    chown $PRODUCT_NAME /home/$PRODUCT_NAME/.cache
+
+USER $PRODUCT_NAME
+ENTRYPOINT ["/bin/nomad-pack"] CMD ["help"]
```
lgfa29, line 52:
```suggestion
ENTRYPOINT ["/bin/nomad-pack"] 
CMD ["help"]
```

I don't know if this was a missing line break or a syntax that I haven't seen before, so feel free to ignore it.

.github/workflows/build.yml
```diff
@@ -0,0 +1,268 @@
...
+        with:
+          go-version: ${{ needs.get-go-version.outputs.go-version }}
+
+      - name: Build
+        env:
+          GOOS: ${{ matrix.goos }}
+          GOARCH: ${{ matrix.goarch }}
+        run: |
+          mkdir dist out
+          go build -o dist/ .
```
lgfa29, line 100:
```suggestion
          make bin
```

I think this would work? Not really necessary though. Maybe standarzing around the same way that [Nomad does it](https://github.com/hashicorp/nomad/blob/main/GNUmakefile#L85-L97) could be helpful for consistency, but probably something for the future.

lgfa29 (approved):
:shipit:

jazzyfresh (approved):
Nice one~!

.release/ci.hcl
angrycub: Hmm.  There isn't one, but not a bad idea to bring one into the pipeline soon.  I'll make an issue to track that.

PR #258, opened 2022-06-04
Add dependency downloading and fix type map of map

piotmni:
Hi, when i was playing with nomad-pack i noticed that there is not code of getting dependency so i tried added it. Also i had a problem when variable type was map(map(..)) because it was treated as a collection of maps so i added a fix for that. What do you think ?


lgfa29 (changes requested):
Hi @piotmni 👋 

Thanks for the PR! I pulling dependencies is a neat idea, but I will check with the rest of the team to get their input as well.

In the meantime, would you mind splitting the map bug fix into a separate PR? Since they are doing two different tasks it will be better to have them as separate work streams as well.

Could you also add some test cases for both? 

For the dependency retrieval work, I think you can just spin-up an [`httptest.Sever`](https://pkg.go.dev/net/http/httptest#Server) and validate that the expected request is made.

Thank you 🙂

piotmni:
Hi @lgfa29 thanks for answer, i will close this PR and continue in another.

PR #264, opened 2022-06-26
helper: swap strings.Title for helper.Title

shoenig:
In Go 1.18+, `strings.Title` is deprecated. This causes `make dev` to fail because
the deprecation is caught by `staticcheck`, e.g.

```
➜ go version 
go version go1.18.3 linux/amd64
```

```
==> Linting source code...
terminal/glint.go:358:37: SA1019: strings.Title is deprecated: The rule Title uses for word boundaries does not handle Unicode punctuation properly. Use golang.org/x/text/cases instead. (staticcheck)
			glint.Text(fmt.Sprintf("! %s\n", strings.Title(sub))),
```

This PR creates `sdk/helper.Title`, which under the hood uses https://pkg.go.dev/golang.org/x/text/cases#Caser
to perform the same function.

More info in https://stackoverflow.com/questions/71620717/in-go-1-18-strings-title-is-deprecated-what-to-use-now-and-how

### DIFF
--- a/internal/cli/stop.go
+++ b/internal/cli/stop.go
@@ -2,7 +2,6 @@ package cli
 
 import (
 	"fmt"
-	"strings"
 	"time"
 
 	v1client "github.com/hashicorp/nomad-openapi/clients/go/v1"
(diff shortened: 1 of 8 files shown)


DerekStrickland (approved):
Thank you!

PR #272, opened 2022-07-18
Fix map of maps variable type

piotmni:
Hi, nomad-pack has problem with variable type map(map(..)). I tried add fix for that and tests for convertCtyToInterface


piotmni:
This PR is separate from https://github.com/hashicorp/nomad-pack/pull/258

lgfa29 (approved):
Thanks for the fix @piotmni!

I pushed a couple of commits to bring the tests closer to how we do them in general (just for style consistency, they worked great 🙂), and also to add a `CHANGELOG` entry.

PR #273, opened 2022-07-18
feat: add downloading dependency

piotmni:
Hi, i tried to add dependency downloading cause currently it is not implemented

### DIFF
--- a/internal/pkg/loader/getter.go
+++ b/internal/pkg/loader/getter.go
@@ -1,0 +1,57 @@
+package loader
+
+
+import (
+	"fmt"
+	"path"
+	"context"
+	"net/http"
+
+	"github.com/hashicorp/go-cleanhttp"
+	"github.com/hashicorp/nomad-pack/sdk/pack"
+	getter "github.com/hashicorp/go-getter"
+)
+
+func GetDependency(dependency *pack.Dependency, depsPath string) error {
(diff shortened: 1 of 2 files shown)


piotmni:
This PR comes from https://github.com/hashicorp/nomad-pack/pull/258 . According to comment, i will try add tests for it later.

lgfa29:
Thanks for splitting the PR @piotmni.

As I mentioned in the previous one, despite this being a pretty straight forward code change, there's a more fundamental design question that we need to discuss, which is how we would handle local dependencies. 

This PR approaches it similarly to vendoring dependencies, so they are co-located with the main project itself, so similar to something like `node_modules` in JavaScript. Another approach would be to read them from a common location, where any missing version would be automatically downloaded there.

We're still not sure which direction we'll go, so  just to set the proper expectations please know that this PR may linger for a while until we reach a conclusion, but we appreciate the contribution!

pkazmierczak:
Hi @piotmni, after some internal discussion we decided to support basic dependency downloading via a new command: `nomad-pack deps vendor`. Please have a look at https://github.com/hashicorp/nomad-pack/pull/367 for context. This functionality will be part of upcoming 0.1 nomad-pack release.

piotmni:
Thx for info

PR #275, opened 2022-07-21
docs: update contributing doc to mention `make boostrap`

lgfa29:
**Description**

`make bootstrap` is a necessary step for development that needs to run at least once.

**Reminders**

- [ ] Add `CHANGELOG.md` entry (not necessary)

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -10,6 +10,8 @@ To add packs, please contribute to the [Nomad Pack Community Registry](https://g
 - Git
 - Make
 
+Make sure you have `$GOHOME/bin` as part of your `$PATH` variable.
+
 ## Building and Running Locally
 
 Installing required tools:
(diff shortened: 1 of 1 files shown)


DerekStrickland (approved):
:shipit:

angrycub:
The `tools` target should be called as a consequence of `make dev` (`dev` -> `lint` -> `tools` -> `lint-deps`).  The `bootstrap` target uses the `tools` target too, it should have been run.  Perhaps we might try a fallback and look in the `$GOHOME/bin` folder in cases where we don't find the tools in the path?

lgfa29:
> The `tools` target should be called as a consequence of `make dev` (`dev` -> `lint` -> `tools` -> `lint-deps`). The `bootstrap` target uses the `tools` target too, it should have been run. Perhaps we might try a fallback and look in the `$GOHOME/bin` folder in cases where we don't find the tools in the path?

The problem is that `tools` take a while to run, even if you already have the tools installed. I'm even considering removing `lint` from `dev` since it also slows down the development flow.

lgfa29:
Updated to mention `$GOHOME/bin`

jrasell (approved):

PR #294, opened 2022-11-02
Update nomad-pack dependencies

angrycub:
**Description**
- Bump .go-version
- Update comments for new gofmt rules
- Update go.mod version to 1.19
- Update nomad 1.3.5 => 1.3.7
- Bump hashicorp/hcl/v2 to v2.9.2-0.20220525143345-ab3cae0737bc
- Update Nomad dep to 1.4.2
- Update nomad-openapi dependency

**Reminders**


lgfa29 (approved):
Thanks!

I pushed a couple of commits to update the test and lint workflow to read the Go version from `.go-version` instead of hardcode them in the workflow.

The linter failures seem legit, although not very helpful as it doesn't point to which files are wrong 😅 

But we can tackle them in a follow-up PR 👍

PR #298, opened 2023-02-23
updates docs links to use `developer.hashicorp.com`

ksatirli:
**Description**

This PR updates all docs links to use `developer.hashicorp.com`

**Reminders**


docs/detailed-usage.md
```diff
@@ -156,7 +156,7 @@ nomad-pack info hello-world
 
 If you do not want to immediately deploy the pack, but instead want details on how it will be deployed, run the `plan` command.
 
-This invokes Nomad in a dry-run mode using the [Nomad Plan](https://www.nomadproject.io/api-docs/jobs#create-job-plan) API endpoint.
+This invokes Nomad in a dry-run mode using the [Nomad Plan](https://developer.hashicorp.com/nomad/api-docs/jobs#create-job-plan) API endpoint.
```
lgfa29, line 159: Is devDot what we should be using now? I feel like www.nomadproject.io is a more stable URL that is unlikely to change in the long-term 🤔

ksatirli: Both are acceptable but I've seen more and more products bring in developer.hashicorp.com (and away from learn.hashicorp.com).

See Vault for [`developer.`](https://github.com/search?q=repo%3Ahashicorp%2Fvault%20developer.hashicorp.com&type=code) and [`.io`](https://github.com/search?q=repo%3Ahashicorp%2Fvault+vaultproject.io&type=code) respectively.

URLs have redirects, but since we - essentially migrated "Learn" to "Developer", I feel that the docs should point to the correct location, too.

lgfa29: Ops, sorry for the delay here. For "Learn" I think it's OK, my point is more about www.nomadproject.io. I don't expect Vault to reference www.vaultproject.io because for them that's just `/`. The same applies for [Nomad](https://github.com/search?q=repo%3Ahashicorp%2Fnomad+nomadproject.io&type=code).

But since this is an external project it needs the full URL, and I think it would be better to keep Nomad docs URLs as www.nomadproject.io

ksatirli: I see both sides of this -

With Vault specifically, you can see some of the [docs](https://github.com/search?q=repo%3Ahashicorp%2Fvault%20developer.hashicorp.com&type=code) links have been moved to `developer.`

Curious if @hashicorp/technical-community-managers has any guideline on how we should treat docs links (and if we have guidance across products)

hsimon-hashicorp: @ksatirli I would personally defer to the docs team on these sorts of issues - for Vault, that's [@yhyakuna](https://github.com/yhyakuna)

PR #303, opened 2023-03-10
renderer: render auxiliary files, and not just templates

### LINKED ISSUE #299: Add support for rendering non-job files
Sometimes a pack might want to include a file that is not a job as part of the render process. This could be a Nomad Policy or Volume that could then be manually used, or something like a config file for later. We should support rendering files like this.

Acceptance Criteria:
- Pack writers can include non-job templates alongside jobspecs in the /templates directory of a pack
- These files will be rendered when the “render” function is called and output alongside rendered jobspecs
- The plan and run commands do not attempt to plan or run these files
- Any file that ends in “.tpl” is rendered

pkazmierczak:
**Description**

This PR allows for rendering any files found in the `templates/` directory, and introduces `--render-aux-files` flag for the `render` command to control that behavior. Resolves #299 

**Reminders**

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -69,7 +69,7 @@ hclfmt: ## Format HCL files with hclfmt
 
 .PHONY: dev
 dev: GOPATH=$(shell go env GOPATH)
-dev: lint
+dev:
 	@echo "==> Building nomad-pack..."
 	@CGO_ENABLED=0 go build -ldflags $(GO_LDFLAGS) -o ./bin/nomad-pack
 	@rm -f $(GOPATH)/bin/nomad-pack
--- a/internal/pkg/loader/loader.go
+++ b/internal/pkg/loader/loader.go
@@ -103,6 +103,12 @@ func loadFiles(files []*pack.File) (*pack.Pack, error) {
 			// object templates and helpers.
 			p.TemplateFiles = append(p.TemplateFiles, f)
 
+		case strings.HasPrefix(f.Name, "templates/") &&
+			!strings.HasSuffix(f.Name, ".tpl"):
+			// if there are any other files inside the "templates/" directory,
+			// add them to the aux files array
+			p.AuxiliaryFiles = append(p.AuxiliaryFiles, f)
+
 		default:
 			// Do nothing with other files; this might change in the future.
 			continue
--- a/internal/pkg/loader/loader.go
+++ b/internal/pkg/loader/loader.go
@@ -104,7 +104,7 @@ func loadFiles(files []*pack.File) (*pack.Pack, error) {
 			p.TemplateFiles = append(p.TemplateFiles, f)
 
 		case strings.HasPrefix(f.Name, "templates/") &&
-			!strings.HasSuffix(f.Name, ".tpl"):
+			strings.HasSuffix(f.Name, ".tpl"):
 			// if there are any other files inside the "templates/" directory,
 			// add them to the aux files array
 			p.AuxiliaryFiles = append(p.AuxiliaryFiles, f)
--- a/internal/pkg/manager/manager.go
+++ b/internal/pkg/manager/manager.go
@@ -113,8 +113,7 @@ func (pm *PackManager) ProcessTemplates(renderAux bool) (*renderer.Rendered, []*
 }
 
 // ProcessOutputTemplate performs the output template rendering.
-func (pm *PackManager) ProcessOutputTemplate(aux bool) (string, error) {
-	pm.renderer.RenderAuxFiles = aux
+func (pm *PackManager) ProcessOutputTemplate() (string, error) {
 	return pm.renderer.RenderOutput()
 }
 
--- a/internal/pkg/manager/manager.go
+++ b/internal/pkg/manager/manager.go
@@ -47,7 +47,7 @@ func NewPackManager(cfg *Config, client *v1.Client) *PackManager {
 // TODO(jrasell) figure out whether we want an error or hcl.Diagnostics return
 // object. If we stick to an error, then we need to come up with a way of
 // nicely formatting them.
-func (pm *PackManager) ProcessTemplates() (*renderer.Rendered, []*errors.WrappedUIContext) {
+func (pm *PackManager) ProcessTemplates(renderAux bool) (*renderer.Rendered, []*errors.WrappedUIContext) {
 
 	loadedPack, err := pm.loadAndValidatePacks()
 	if err != nil {
@@ -98,6 +98,9 @@ func (pm *PackManager) ProcessTemplates() (*renderer.Rendered, []*errors.Wrapped
 	r.Client = pm.client
 	pm.renderer = r
 
+	// should auxiliary files be rendered as well?
+	pm.renderer.RenderAuxFiles = renderAux
+
 	rendered, err := r.Render(loadedPack, mapVars)
 	if err != nil {
 		return nil, []*errors.WrappedUIContext{{
@@ -110,7 +113,10 @@ func (pm *PackManager) ProcessTemplates() (*renderer.Rendered, []*errors.Wrapped
 }
 
 // ProcessOutputTemplate performs the output template rendering.
-func (pm *PackManager) ProcessOutputTemplate() (string, error) { return pm.renderer.RenderOutput() }
+func (pm *PackManager) ProcessOutputTemplate(aux bool) (string, error) {
+	pm.renderer.RenderAuxFiles = aux
+	return pm.renderer.RenderOutput()
+}
 
 // loadAndValidatePacks triggers the initial parent load and then starts the
 // dependent pack loader. The returned pack will therefore be fully populated.
--- a/internal/pkg/renderer/renderer.go
+++ b/internal/pkg/renderer/renderer.go
@@ -46,9 +46,11 @@ const (
 // template using the parsed variable map.
 func (r *Renderer) Render(p *pack.Pack, variables map[string]interface{}) (*Rendered, error) {
 
-	// templatesToRender stores all the template that should be rendered.
+	// templatesToRender stores all the template that should be rendered,
+	// auxFilesToRender stores all the auxiliary files that should be rendered.
 	templatesToRender := make(map[string]toRender)
-	prepareTemplates(p, templatesToRender, variables)
+	auxFilesToRender := make(map[string]toRender)
+	prepareFiles(p, templatesToRender, auxFilesToRender, variables)
 
 	// Set up our new template, add the function mapping, and set the
 	// delimiters.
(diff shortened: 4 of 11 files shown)


internal/pkg/renderer/renderer.go
```diff
@@ -164,13 +173,20 @@ func prepareTemplates(p *pack.Pack, templates map[string]toRender, variables map
...
 
 	// Add each template within the pack with scoped variables.
 	for _, t := range p.TemplateFiles {
 		templates[path.Join(p.Name(), t.Name)] = toRender{content: string(t.Content), variables: newVars}
 	}
+
+	if renderAuxFiles {
+		// Add each aux file within the pack with scoped variables.
+		for _, f := range p.AuxiliaryFiles {
+			templates[path.Join(p.Name(), f.Name)] = toRender{content: string(f.Content), variables: newVars}
```
lgfa29, line 187:
```suggestion
			auxFiles[path.Join(p.Name(), f.Name)] = toRender{content: string(f.Content), variables: newVars}
```

I think? Otherwise `auxFiles` doesn't seem to be used anywhere. But then I think `Renderer.Render` needs to handle this map, which it doesn't seem to be doing.

More general, it may be worth refactoring this function so it doesn't take maps as inputs, but rather return them.

internal/pkg/renderer/renderer.go
```diff
@@ -136,10 +142,13 @@ func (r *Renderer) RenderOutput() (string, error) {
...
-// passed map with the templates to render along with the variables which
-// correspond.
-func prepareTemplates(p *pack.Pack, templates map[string]toRender, variables map[string]interface{}) {
+// prepareFiles recurses the pack and its dependencies to populate to the passed
+// map with the templates to render along with the variables which correspond.
+func prepareFiles(p *pack.Pack,
+	templates map[string]toRender,
+	auxFiles map[string]toRender,
+	variables map[string]interface{},
+	renderAuxFiles bool) {
```
lgfa29, line 151:
```suggestion
	renderAuxFiles bool,
) {
```

Just to keep it consistent with other line breaks.

internal/pkg/loader/loader.go
```diff
@@ -103,6 +103,12 @@ func loadFiles(files []*pack.File) (*pack.Pack, error) {
 			// object templates and helpers.
 			p.TemplateFiles = append(p.TemplateFiles, f)
 
+		case strings.HasPrefix(f.Name, "templates/") &&
+			!strings.HasSuffix(f.Name, ".tpl"):
```
lgfa29, line 107: Do you have an example of a use case for this feature? Would it be possible to just change the `case` above to something like this?

```diff
- strings.HasSuffix(f.Name, ".nomad.tpl") ||
+ strings.HasSuffix(f.Name, ".tpl") ||
```

lgfa29 (changes requested):
Maybe I'm holding it wrong, but it seems like the `auxFiles` handling is wrong, meaning that auxiliary files are treated just like templates, in which case we may not need a new config at all?

Not part of this PR, but it would also be nice to sort the files so they are always rendered in the same order. Right now the output order is random.

pkazmierczak:
> Maybe I'm holding it wrong, but it seems like the `auxFiles` handling is wrong, meaning that auxiliary files are treated just like templates, in which case we may not need a new config at all?

You are _not_, in fact, holding it wrong! Mistakes were made, I refactored in 6151776461dc689ac79b6dd45011e88b8103c912. Now there's just one map, and now the helper function `prepareFiles` returns it, doesn't take it as an argument.

internal/pkg/renderer/renderer.go
pkazmierczak: Addressed in 6151776461dc689ac79b6dd45011e88b8103c912

internal/pkg/loader/loader.go
pkazmierczak: Here's how I understood #299: we want files other than `.tpl` that could be found inside `templates/` to be rendered. Does that make sense? I can't really think of a usecase, I'm still pretty new to pack.

pkazmierczak: #299 says: 

> Pack writers can include non-job templates alongside jobspecs in the /templates directory of a pack

So I guess we assume these would also have `.tpl` extension?

lgfa29: Yeah, I think I'm reading that the same way as @angrycub. The last acceptance criteria mentions that:

> Any file that ends in “.tpl” is rendered

I think we can even avoid making assumptions about file content based on extenstion? For example, we could try to parse files as job and, if that fails, then we know that the file is not a job. Maybe it's a volume? So try to parse it as a volume etc.

This goes back to the discussion of having better scoped entities in Nomad, but that's a different discussion 😄

pkazmierczak: I think I'm still a bit confused. I changed the conditional in 99f3544, let me know if this is what we want.

internal/pkg/manager/manager.go
```diff
@@ -110,7 +113,10 @@ func (pm *PackManager) ProcessTemplates() (*renderer.Rendered, []*errors.Wrapped
 }
 
 // ProcessOutputTemplate performs the output template rendering.
-func (pm *PackManager) ProcessOutputTemplate() (string, error) { return pm.renderer.RenderOutput() }
+func (pm *PackManager) ProcessOutputTemplate(aux bool) (string, error) {
+	pm.renderer.RenderAuxFiles = aux
```
lgfa29, line 117:
```suggestion
func (pm *PackManager) ProcessOutputTemplate() (string, error) {
```

I'm still getting used to the code base and the lifecylce of each component, but it seems like `ProcessOutputTemplate` is expected to only be called after `ProcessTemplates` (otherwise `pm.renderer` is `nil`), and so `RenderAuxFiles` should also be expected to be set?

internal/pkg/renderer/renderer.go
```diff
@@ -119,8 +124,8 @@ func (r *Renderer) Render(p *pack.Pack, variables map[string]interface{}) (*Rend
 // RenderOutput performs the output template rendering.
 func (r *Renderer) RenderOutput() (string, error) {
 
-	// If we don't have a template file, then return early.
-	if r.pack.OutputTemplateFile == nil {
+	// If we don't have a template file or any aux files, return early.
+	if r.pack.OutputTemplateFile == nil && r.pack.AuxiliaryFiles == nil {
```
lgfa29, line 128:
```suggestion
	if r.pack.OutputTemplateFile == nil && len(r.pack.AuxiliaryFiles) == 0 {
```

I think an empty slice is not `nil`. But also, it doesn't seem like `r.pack.AuxiliaryFiles` is used here, so maybe we don't need this check?

lgfa29 (approved):
Some other minor comments, but LGTM. 

I also noticed a bug where `TrimRight` is used instead of `TrimPrefix` when [generating the file name](https://github.com/hashicorp/nomad-pack/blob/29712e33b0f784922fa384857934be87a6ac8251/internal/cli/render.go#L152). So if you have a file called `test.hcl.tpl` it will output `test.hc`. [`cloneRemoteGitRegistry`](https://github.com/hashicorp/nomad-pack/blob/main/internal/pkg/cache/add.go#L136) has a similar problem. I don't know if you want to go ahead and fix those here of if I should open a follow-up issue about them.

GNUmakefile
```diff
@@ -69,7 +69,7 @@ hclfmt: ## Format HCL files with hclfmt
 
 .PHONY: dev
 dev: GOPATH=$(shell go env GOPATH)
-dev: lint
+dev:
```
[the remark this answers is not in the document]
pkazmierczak: yeah, it takes a lot of time for `dev` builds. I think it's okay to disable it for dev.

internal/pkg/manager/manager.go
pkazmierczak: You are right of course! Fixed in c273f650df96ecd57b64bfe0486846badebb4466

pkazmierczak:
> I also noticed a bug where `TrimRight` is used instead of `TrimPrefix` when [generating the file name](https://github.com/hashicorp/nomad-pack/blob/29712e33b0f784922fa384857934be87a6ac8251/internal/cli/render.go#L152). So if you have a file called `test.hcl.tpl` it will output `test.hc`. [`cloneRemoteGitRegistry`](https://github.com/hashicorp/nomad-pack/blob/main/internal/pkg/cache/add.go#L136) has a similar problem. I don't know if you want to go ahead and fix those here of if I should open a follow-up issue about them.

Addressed in https://github.com/hashicorp/nomad-pack/pull/309

internal/pkg/renderer/renderer.go
lgfa29: It does make sense, but adding this check does the opposite of that when `len(r.pack.AuxiliaryFiles) > 0` and `r.pack.OutputTemplateFile == nil`. Nothing inside this function cares about `r.pack.AuxiliaryFiles` so running it just because there are auxiliary files is unnecessary when `r.pack.OutputTemplateFile` is `nil`. 

But worse, looking at it again, this change actually introduces a `nil` panic in this scenario:

```
$ nomad-pack render  --render-output-template .

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x1031e9dac]

goroutine 1 [running]:
github.com/hashicorp/nomad-pack/internal/pkg/renderer.(*Renderer).RenderOutput(0x140006915c0)
        /Users/laoqui/go/src/github.com/hashicorp/nomad-pack/internal/pkg/renderer/renderer.go:132 +0x3c
github.com/hashicorp/nomad-pack/internal/pkg/manager.(*PackManager).ProcessOutputTemplate(...)
        /Users/laoqui/go/src/github.com/hashicorp/nomad-pack/internal/pkg/manager/manager.go:118
github.com/hashicorp/nomad-pack/internal/cli.(*RenderCommand).Run(0x1400058e540, {0x140000300a0?, 0x0?, 0xffffffffffffffff?})
        /Users/laoqui/go/src/github.com/hashicorp/nomad-pack/internal/cli/render.go:232 +0x728
github.com/mitchellh/cli.(*CLI).Run(0x140000003c0)
        /Users/laoqui/go/pkg/mod/github.com/mitchellh/cli@v1.1.4/cli.go:262 +0x4a8
github.com/hashicorp/nomad-pack/internal/cli.Main({0x14000030080?, 0x4, 0x4})
        /Users/laoqui/go/src/github.com/hashicorp/nomad-pack/internal/cli/main.go:93 +0x268
main.main()
        /Users/laoqui/go/src/github.com/hashicorp/nomad-pack/main.go:12 +0x90
```

pkazmierczak: 🤦 https://github.com/hashicorp/nomad-pack/pull/310🤦

Issue #308, opened 2023-03-20
Nomad Pack Beta (meta issue)


mikenomitch:
This is a meta issue for capturing the various things we want to do for the "beta" Pack release.

The contents of this are subject to change as we get further along:

Quality of Life Improvements:

- [ ] https://github.com/hashicorp/nomad-pack/issues/299
- [ ] https://github.com/hashicorp/nomad-pack/issues/188
- [ ] https://github.com/hashicorp/nomad-pack/issues/301
- [ ] https://github.com/hashicorp/nomad-pack/issues/25
- [ ] https://github.com/hashicorp/nomad-pack/issues/302
- [ ] https://github.com/hashicorp/nomad-pack/issues/296
- [ ] https://github.com/hashicorp/nomad-pack/issues/284
- [ ] Add ability to pass in HCL override blocks (to be further spec'd)

Structural Changes:
- [ ] https://github.com/hashicorp/nomad-pack/issues/304
- [ ] https://github.com/hashicorp/nomad-pack/issues/300
- [ ] Improve nested packs (better variable scoping, metadata access)
- [ ] Change "registry list" to list registries and add "list" for listing Packs

Bug Fixes:
- [ ] https://github.com/hashicorp/nomad-pack/issues/279

Outside of the Repo:
- [ ] Add Packs and Registries to Nomad-specific HashiCorp Integrations page (similar to [Vault's](https://developer.hashicorp.com/vault/integrations)).

attachmentgenie:
@mikenomitch i would like to make a case for #153 and #266. These tickets describe IMHO reasonably annoying behavior that cause nomad-pack to behave different from vanilla nomad

david-yu:
@angrycub Should we close this as well?

PR #311, opened 2023-03-29
renderer: format templates before outputting

### LINKED ISSUE #301: Rendered jobs should be auto-formatted
By default, jobs should be formatted with hclfmt. It is difficult to get spacing correct using gotemplate so many of our jobs are unfortunately ugly. This can fix that.

There should be a flag that allows not formatting a pack on render: nomad-pack render <PACK-NAME> –no-format

pkazmierczak:
**Description**

This PR introduces HCL formatting of templates before they are outputted. An additional flag `--no-format` is now also available for the `render` command to override this behavior. 

Resolves #301 
Relates to #308 

**Reminders**


lgfa29 (approved):

slonopotamus:
This doesn't work properly with HCL2 job files. See #420 for a fix.

PR #312, opened 2023-03-29
fix: change render-aux-files flag behavior for consistency

pkazmierczak:
Since we want to render auxiliary files by default, the flag to control it should be `--no-render-aux-files` rather than `--render-aux-files`. See https://github.com/hashicorp/nomad-pack/pull/311 for an example of similar behavior.

### DIFF
--- a/internal/cli/render.go
+++ b/internal/cli/render.go
@@ -290,7 +290,7 @@ func (c *RenderCommand) Flags() *flag.Sets {
 		})
 
 		f.BoolVar(&flag.BoolVar{
-			Name:    "no-render-aux-files",
+			Name:    "skip-aux-files",
 			Target:  &c.noRenderAuxFiles,
 			Default: false,
 			Usage: `Controls whether or not the rendered output contains auxiliary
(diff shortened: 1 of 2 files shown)


internal/cli/render.go
```diff
@@ -290,9 +290,9 @@ func (c *RenderCommand) Flags() *flag.Sets {
 		})
 
 		f.BoolVar(&flag.BoolVar{
-			Name:    "render-aux-files",
-			Target:  &c.renderAuxFiles,
-			Default: true,
+			Name:    "no-render-aux-files",
```
lgfa29, line 293:
```suggestion
			Name:    "skip-aux-files",
```

`--no-render-aux-files` is quite a mouthful 😅 

The command is also `nomad-pack render` already, so that part seems redundant. I also wonder if we need a CLI flag at all? When would we want to _not_ render these files?

lgfa29 (approved):

internal/cli/render.go
pkazmierczak: I guess my idea behind having that flag was to allow users to preserve previous behavior?

PR #314, opened 2023-03-31
admin: adjusted CODEOWNERS file

pkazmierczak:
I imagine it must be getting somewhat annoying that @angrycub @jrasell and @lgfa29 get tagged automatically on every PR. This also removes users that aren't part of hashicorp org anymore.


jrasell (approved):
LGTM, thanks @pkazmierczak!

PR #317, opened 2023-04-04
pack: Author field no longer supported in metadata

### LINKED ISSUE #300: Remove "author" from pack metadata
We should remove the author field from pack metadata completely.

There is consistent confusion whether or not the author means the Pack maintainer or the App author. The pack maintainer is information that can be derived from Git history and often doesn’t have a single true source (as some have multiple community contributors). The app author can be derived from the URL. 

app {
    url = “www.nginx.com”
    author = “Greg Moss”
}

If author is in metadata, ideally it doesn't break the pack, as old packs might still include it. If there is no way to nicely do this, then we should batch this with any other breaking changes to the metadata we make.

pkazmierczak:
**Description**

This PR removes the `Author` field from pack metadata. 

Resolves #300 
Relates to #308 

**Reminders**

### DIFF
--- a/sdk/pack/metadata.go
+++ b/sdk/pack/metadata.go
@@ -24,8 +24,12 @@ type MetadataApp struct {
 	// quick reference to the documentation and help pages.
 	URL string `hcl:"url"`
 
-	// ExtraKeysHCL is used by hcl to surface unexpected keys.
-	ExtraKeysHCL []string `hcl:",unusedKeys" json:"-"`
+	// Author is an identifier to the author and maintainer of the pack such as
+	// HashiCorp or James Rasell
+	//
+	// Deprecated: Nomad Pack tech preview 4 removes this field, we keep it here for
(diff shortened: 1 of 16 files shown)


sdk/pack/metadata.go
```diff
@@ -24,6 +24,7 @@ type MetadataApp struct {
 	// quick reference to the documentation and help pages.
 	URL string `hcl:"url"`
 
+	// OBSOLETED in nomad pack beta!
 	// Author is an identifier to the author and maintainer of the pack such as
 	// HashiCorp or James Rasell
 	Author string `hcl:"author"`
```
angrycub, line 30: Let's use ExtraKeys
```suggestion
	ExtraKeysHCL []string `hcl:",unusedKeys" json:"-"`
```

angrycub (approved):
Not going to block on that change.

PR #318, opened 2023-04-05
registry: store refs and source URLs of cloned repositories

### LINKED ISSUE #170: Registry source URL can not reliably be recreated from Pack.URL scanning
Currently, pack uses a scan of pack metadata in the registry folder to make a best guess as to what the repository path should be. This approach would break in cases where the repository is a fork of a source repo, because the pack URLs will all refer to the upstream rather than the fork itself.

Proposal: Create a metadata file at the repository level that contains the URL that the user used when adding the registry. This file could also eventually contain more user-supplied or even registry-supplied data.  This would be similar to the way in which git tracks remotes in the .git/config file.

### LINKED ISSUE #304: Registry metadata improvements
Current registry implementation is lacking in a few areas, but there is a couple of easy fixes we could make before going full "vault registry" way. 

- Metadata should include a source URL to avoid namespace clashes in case users fork repositories, for example. 
- We should allow users to provide other types of URLs, not just github, but also s3 or local paths, since it's not an issue for go-getter. 
- Metadata should also store the sha of the repository if users just want "latest", so that we could later enable detecting stale packs.

pkazmierczak:
**Description**

This PR adds support for storing refs of cloned registry repositories as well as their full URLs. This way we avoid namespace clashes with forked repos and we'll be able to support "stale registry" detection in the future. It also means that pack no longer relies on `url` metadata property, which can now be safely removed (vide https://github.com/hashicorp/nomad-pack/pull/240). 

Resolves #304 
Resolves #170 
Relates to #308

**Reminders**


internal/pkg/cache/add.go
```diff
@@ -45,13 +47,30 @@ func (c *Cache) Add(opts *AddOpts) (cachedRegistry *Registry, err error) {
...
+		b, err := json.MarshalIndent(cachedRegistry, "", "  ")
+		if err != nil {
+			return cachedRegistry, err
+		}
+
+		metaPath := fmt.Sprintf("%s/%s/metadata.json", c.cfg.Path, opts.RegistryName)
+		err = os.WriteFile(metaPath, b, 0644)
+		if err != nil {
+			return cachedRegistry, err
+		}
```
jrasell, line 70: This could use less lines, although this is personal preference and not a required change.
```suggestion
		if err = os.WriteFile(metaPath, b, 0644); err != nil {
			return cachedRegistry, err
		}
```

internal/pkg/cache/add.go
```diff
@@ -153,15 +176,21 @@ func (c *Cache) cloneRemoteGitRegistry(opts *AddOpts) (err error) {
 	if opts.PackName != "" {
 		clonePath = path.Join(clonePath, "packs", opts.PackName)
 	}
-	err = gg.Get(clonePath, fmt.Sprintf("git::%s", url))
+	err := gg.Get(clonePath, fmt.Sprintf("git::%s", url))
 	if err != nil {
 		logger.ErrorWithContext(err, "could not install registry", c.ErrorContext.GetAll()...)
-		return
+		return "n/a", err
+	}
```
jrasell, line 183: We could shorten this slightly, but again not required and personal style. Do we need to document how to handle "n/a" and use a const, or is this thinking too much and not required?
```
	if err := gg.Get(clonePath, fmt.Sprintf("git::%s", url)); err != nil {
		logger.ErrorWithContext(err, "could not install registry", c.ErrorContext.GetAll()...)
		return "n/a", err
	}
```

internal/pkg/cache/cache.go
```diff
@@ -213,3 +217,21 @@ func refFromPackEntry(packEntry os.DirEntry) (ref string) {
 
 	return
 }
+
+// getGitHeadRef is a helper method that takes a directory which is a git
+// repository, and returns the SHA of the git HEAD of that repository.
+func getGitHeadRef(clonePath string) (string, error) {
+	sha := "n/a"
```
jrasell, line 224: Same comment as within the `add.go` file.

internal/pkg/cache/cache.go
```diff
@@ -213,3 +217,21 @@ func refFromPackEntry(packEntry os.DirEntry) (ref string) {
...
+	r, err := git.PlainOpen(clonePath)
+	if err != nil {
+		return sha, fmt.Errorf("could not read cloned repository: %v", err)
+	}
+	if r != nil {
```
jrasell, line 229: Can r be nil if an error is returned above?

internal/pkg/cache/registry.go
```diff
@@ -48,6 +55,22 @@ func (r *Registry) get(opts *GetOpts, cache *Cache) (err error) {
 			continue
 		}
 
+		// Read the top-level metadata file but don't process it like a pack
+		if packEntry.Name() == "metadata.json" {
+			f, err2 := os.ReadFile(path.Join(opts.RegistryPath(), packEntry.Name()))
```
jrasell, line 60: Curious why we are using err2 here and not err?

jrasell (approved):
LGTM, some minor comments and questions but nothing blocking.

internal/pkg/cache/add.go
```diff
@@ -45,13 +46,28 @@ func (c *Cache) Add(opts *AddOpts) (cachedRegistry *Registry, err error) {
...
 
-	return
+	if cachedRegistry != nil {
+		// Store a metadata JSON file for the cached registry
+		b, err := json.MarshalIndent(cachedRegistry, "", "  ")
+		if err != nil {
+			return cachedRegistry, err
+		}
+
+		metaPath := fmt.Sprintf("%s/%s/metadata.json", c.cfg.Path, opts.RegistryName)
```
shoenig, line 64:
```suggestion
		metaPath := fmt.Sprintf("%s/%s/metadata.json", c.cfg.Path, opts.RegistryName)
```

use `filepath.Join` for safety and cross-platform

shoenig (approved):
Sorry I missed this @pkazmierczak, just the one thing from me

internal/pkg/cache/cache.go
pkazmierczak: fair point!

internal/pkg/cache/registry.go
pkazmierczak: if it's an `err`, linter complains that it shadows previous err declaration.

PR #325, opened 2023-04-24
[chore] 1.19 go mod tidy

angrycub:
Looks like we missed a `go mod tidy` in a recent PR. This adds it for the 1.19 go version. Alternatively, we could 
use PR #324 and get its tidied go.mod as a side-effect.


pkazmierczak (approved):

angrycub:
Closed with #324

PR #349, opened 2023-05-22
Migrate CRT configuration to 'prepare' workflow

smacfarlane:
This replaces the explicit configuration enumerating each step in the common release tooling process with a single step that encapsulates all organizational requirements (signing, notarization, etc.).

### DIFF
--- a/.release/ci.hcl
+++ b/.release/ci.hcl
@@ -18,173 +18,19 @@ project "nomad-pack" {
   }
 }
 
-event "merge" {
-  // "entrypoint" to use if build is not run automatically
-  // i.e. send "merge" complete signal to orchestrator to trigger build
-}
+event "build" {}
 
-event "build" {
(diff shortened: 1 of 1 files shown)


pkazmierczak:
Hi @smacfarlane, thanks for the PR! Can we get a bit more context, though? Is the `build` step intentionally left empty?

smacfarlane:
Hi @pkazmierczak ,  yes the `build` is intentionally empty.  Due to an implementation quirk, it's needed as a stub entry, but the contents of the block are never used. 

This is part of an effort on Release Engineering to simplify the required configuration for the common release tooling, and decrease the time it takes to make artifacts ready for release. We would like to disable the single-purpose workflows ( ex: `notarize-darwin-amd64`), but migration to `prepare` is a requirement for that.

pkazmierczak (approved):

PR #351, opened 2023-05-23
Integration stanza support

BrandonRomano:
**Description**
Adds in support for an integration stanza, as to allow for the registration of nomad packs into our upcoming Nomad integrations library (ex: [waypoint's library](https://developer.hashicorp.com/waypoint/integrations))

### DIFF
--- a/sdk/pack/metadata.go
+++ b/sdk/pack/metadata.go
@@ -101,6 +101,11 @@ func (md *Metadata) ConvertToMapInterface() map[string]interface{} {
 				"description": md.Pack.Description,
 				"version":     md.Pack.Version,
 			},
+			"integration": map[string]interface{}{
+				"identifier": md.Integration.Identifier,
+				"flags":      md.Integration.Flags,
+				"name":       md.Integration.Name,
+			},
 		},
 	}
 }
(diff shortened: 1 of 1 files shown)


pkazmierczak:
Hey @BrandonRomano, thanks a lot for the PR! This looks good, but could you adjust our unit tests in `metadata_test.go`? If you're busy, we can merge this and I can do the rest myself. (can't push commits since it's your fork and not a local branch)

sdk/pack/metadata.go
```diff
@@ -11,9 +11,10 @@ import (
 // high-level information about the pack which is useful for operators and is
 // also exposed as template variables during rendering.
 type Metadata struct {
-	App          *MetadataApp  `hcl:"app,block"`
-	Pack         *MetadataPack `hcl:"pack,block"`
-	Dependencies []*Dependency `hcl:"dependency,block"`
+	App          *MetadataApp         `hcl:"app,block"`
+	Pack         *MetadataPack        `hcl:"pack,block"`
+	Integration  *MetadataIntegration `hcl:"integration,block,optional"`
```
pkazmierczak, line 16: I don't think a `block` can be `optional`. I think if you want to have that block optional, just mark all the fields in `MetadataIntegration` as optional instead.

BrandonRomano:
@pkazmierczak ; I can give it a shot shortly.  I'll let you know if I find myself without enough time to do that.

I'll work to get permissions on this repo so my next PR can be just a branch, if needed!

BrandonRomano:
Okey @pkazmierczak, I've patched up the tests in `metadata_test.go`.

When I run the tests though, other tests appear to be failing due to some null pointer references, I think being caused by `ConvertToMapInterface` / `AddToInterfaceMap` referencing the Integration's values (and Integration is nil).

I think if we have to have the `Integration` object as non optional, we'll run into this / fix this up in other places.  I think if it can be optional, we can just update the `AddToInterfaceMap` / `ConvertToMapInterface` functions to have the nil check in place. Not sure though, what is most appropriate.

I think this is all I can dedicate for now, so if you want to merge / take over, feel free! Happy to continue collaborating later today after some meetings.

PR #352, opened 2023-05-25
refactor: replace nomad-openapi with official nomad API

pkazmierczak:
This PR removes nomad-openapi in favor of Nomad's official API, and updates all the dependencies.


mikenomitch (approved):
No opposition to the OpenAPI removal from me

internal/cli/helpers.go
```diff
@@ -387,97 +387,102 @@ func getDeployedPackJobs(c *v1.Client, cfg *cache.PackConfig, deploymentName str
...
+// address otherwise.
+func handleBasicAuth(s string) (string, string, string) {
+	before, after, found := strings.Cut(s, "@")
+	if found {
+		user, pass, found := strings.Cut(before, ":")
+		if found {
+			return user, pass, after
+		}
+	}
+	return "", "", before
```
shoenig, line 414: I was thinking surely there must be a standard library function for this, but I don't see one :thinking: 

The closes is https://cs.opensource.google/go/go/+/refs/tags/go1.20.4:src/net/http/request.go;l=949 which is about extracting from the Authorization header which isn't the same.

shoenig (approved):

internal/cli/helpers.go
pkazmierczak: yup, went through the same reasoning. Sadly it is what it is.

PR #356, opened 2023-06-02
cache: allow for different types of caches and different versions of registries at the same time

pkazmierczak:
**Description**

This PR changes the way registries are stored within the cache, and thus allows for having multiple versions of the same registries stored without conflict. 

Previous scheme was: `registryName/packName@ref`
New scheme is: `registryName/registryRef/packName@ref`

**Reminders**


shoenig (approved):
LGTM!

PR #358, opened 2023-06-14
[bug] Missing nil check for err when PackPath exists.

angrycub:
**Description**
When the pack path exists err is nil, however we didn't check for that causing us to slide into the error pathway 
and return a nil error that makes other things sad later.

**Reminders**

- [ ] ~Add `CHANGELOG.md` entry~ Since this bug never shipped in a versioned release, will omit CL entry


pkazmierczak (approved):
Nice catch, thanks Charlie.

PR #366, opened 2023-06-26
template: check namespace and region before submitting templates to the client

### LINKED ISSUE #364: Permission denied to plan and run a job with namespace using client type token
### Nomad version
Nomad v1.5.6

### Nomad-pack version
Nomad Pack v0.0.1-techpreview.4 (50d84d0)

### Issue

Nomad-pack fails to plan and run a job with a namespace when it uses a client type token, using the same token with nomad it's possible to execute plan and run.

### Reproduction steps

Create a policy with the following capabilities, in this example the namespace is `testns`

test-policy.hcl
```
namespace "testns" {
  capabilities = [
    "list-jobs",
    "parse-job",
    "read-job",
    "submit-job",
    "read-logs",
    "alloc-exec",
    "alloc-lifecycle"
  ]
}
```

`nomad acl policy apply test-policy test-policy.hcl`

Create a token and associate the previous policy

`nomad acl token create -name 'test-token' -policy test-policy`

load the newly created token

`export NOMAD_TOKEN=xxxxxxxxx`

plan or run a job with nomad-pack

`/tmp/nomad-pack plan hello_world --var register_consul_service=false --var count=1`

#### Expected Result

`/tmp/nomad-pack plan hello_world --var register_consul_service=false --var count=1`

```
+ Job: "hello_world"
+ Task Group: "app" (1 create)
  + Task: "server" (forces create)

» Scheduler dry-run:
Plan succeeded
```

#### Actual Result

```
$ nomad-pack plan hello_world --var register_consul_service=false --var count=1
! Failed To Parse Job Specification

	Error:   Permission denied
	Type:    *v1.APIError
	Context:
	         - Template Name: hello_world/templates/hello_world.nomad.tpl
	         - Registry Name: default
	         - Pack Name: hello_world
	         - Pack Ref: latest
	         - Deployment Name: hello_world@latest
```

### Job file

```bash
$ /tmp/nomad-pack render hello_world --var register_consul_service=false --var count=1
hello_world/hello_world.nomad:

job "hello_world" {
  datacenters = ["<redacted>"]
  namespace   = "testns"
  type        = "service"

  group "app" {
    count = 1

    network {
      port "http" {
        to = 8000
      }
    }

    restart {
      attempts = 2
      interval = "30m"
      delay    = "15s"
      mode     = "fail"
    }

    task "server" {
      driver = "docker"

      config {
        image = "mnomitch/hello_world_server"
        ports = ["http"]
      }

      env {
        MESSAGE = "Hello World!"
      }
    }
  }
}
```

### Nomad Server logs 

[DEBUG] http: request failed: method=POST path=/v1/jobs/parse error="Permission denied" code=403
[DEBUG] http: request complete: method=POST path=/v1/jobs/parse duration=1.501077ms

pkazmierczak:
**Description**

If Nomad uses ACLs that restrict our access by namespace or region, we need to extract namespace/region information from the raw templates before submitting them to the client; otherwise we get a 403. 

Resolves #364

**Reminders**


jrasell (approved):
LGTM.

pschichtel:
@pkazmierczak I'm not sure I follow the code correctly, but it seems to apply the regexes to the templates before rendering. is that correct? and if so, what if the region is based on a template parameter like `region = [[ .my.some_var | quote ]]`?

slonopotamus:
This doesn't work properly. My job has

```hcl
variable "regions" {
  type = list(object({
    region = string
    count  = number
  }))
  default = [
    {
      region = "EU"
      count  = 1
    }
  ]
  description = "Specifies dedicated server instance count per region"
}
```

This variable is _not_ a nomad region, it is just a random variable. However, nomad-pack thinks that

1. This is a Nomad region
2. It tries to apply this region to all dependent pack jobs too, which is completely wrong because different jobs from different packs might be deployed to different regions.

Obviously, `nomad-pack run` fails with `No path to region` error while `nomad-pack render` + `nomad run` just works.

So, this regex-based approach is fundamentally broken because it picks up completely unrelated strings from job file. I suggest reverting this change.

PR #367, opened 2023-06-28
pack: vendor dependencies

pkazmierczak:
**Description**

This PR adds a new CLI command: `deps vendor` that automatically downloads dependencies found in current pack's `metadata.hcl` and stores them inside a `deps/` subdirectory.  The command supports the following options:
```
      --path=<string>
        Full path to the pack which contains dependencies to be vendored. All
        the dependencies will then be downloaded into a 'deps/' subdirectory
        of that path.

      --timeout=<int> (default 30)
        Timeout (in seconds) for downloading dependencies. Defaults to 30.
```

An example user workflow would be:

```
~/my_pack $ nomad-pack deps vendor
  downloading hello_pack pack to vendor/hello_pack...
...success!
  downloading fabio_pack pack to vendor/fabio_pack...
...success!
```

In order for `deps vendor` command to work properly, users need to specify their dependencies in the pack metadata.hcl, with `source` pointing to where the dependency is, and name label matching the dependent pack name, as specified in its metadata. E.g.:
```
app {
  url = "https://some-url-for-the-application.dev"
}

pack {
  name = "simple_service"
  description = "This pack contains a simple service job, and depends on another pack."
  version = "0.2.1"
}

dependency "demo_dep" {
  source = “git://source.git/packs/demo_dep”
}
```

**Reminders**


internal/cli/main.go
```diff
@@ -10,10 +10,12 @@ import (
 	"sort"
 	"text/tabwriter"
 
-	flag "github.com/hashicorp/nomad-pack/internal/pkg/flag"
-	"github.com/hashicorp/nomad-pack/internal/pkg/version"
 	"github.com/mitchellh/cli"
 	"github.com/mitchellh/go-glint"
+
+	flag "github.com/hashicorp/nomad-pack/internal/pkg/flag"
```
jrasell, line 16:
```suggestion
	"github.com/hashicorp/nomad-pack/internal/pkg/flag"
```

internal/pkg/cache/add.go
```diff
@@ -327,6 +327,8 @@ type AddOpts struct {
 	Username string
 	// Optional password for basic auth to a registry that requires authentication.
 	Password string
+	// Temporary pack location that we want to copy it from.
+	CurrentLocation string
```
jrasell, line 331: I know the other parameters also do this, but comments should start with the parameter name.

internal/pkg/deps/vendor.go
```diff
@@ -0,0 +1,43 @@
...
+	// attempt to read metadata.hcl
+	metadata := &pack.Metadata{}
+	err := hclsimple.DecodeFile(path.Join(targetPath, "metadata.hcl"), nil, metadata)
+	if err != nil {
+		return err
+	}
+
+	if len(metadata.Dependencies) == 0 {
+		return fmt.Errorf("metadata.hcl file does not contain any dependencies")
+	}
```
jrasell, line 30: If someone runs the command on a pack with zero dependencies this will return an error? Did we consider this being a noop, or just an info log message?

internal/pkg/helper/signalcontext.go
```diff
@@ -1,7 +1,4 @@
-// Copyright (c) HashiCorp, Inc.
-// SPDX-License-Identifier: MPL-2.0
```
jrasell, line 2: Should we keep this?

internal/pkg/cache/add.go
pkazmierczak: Good catch, but this field is actually obsolete, it's not needed since we don't support copying to cache anymore. I'll remove it.

internal/pkg/helper/signalcontext.go
pkazmierczak: oof we really should! fixed.

internal/pkg/deps/vendor.go
pkazmierczak: I would consider it an error. The command is specifically intended to download dependencies, i.e., my understanding is that if you run it as a user, you expect the `metadata.hcl` to contain dependency information. If it doesn't contain it, we should return an error.

PR #373, opened 2023-07-07
cli: allow filtering packs by registry

pkazmierczak:
**Description**

This PR adds a `--registry` flag to the `nomad-pack list` command, which allows for filtering packs by registry. 

**Reminders**

### DIFF
--- a/internal/cli/list.go
+++ b/internal/cli/list.go
@@ -14,6 +14,7 @@ import (
 // to the current machine.
 type ListCommand struct {
 	*baseCommand
+	registry string
 }
 
 func (c *ListCommand) Run(args []string) int {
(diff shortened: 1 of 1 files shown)


pkazmierczak:
@philrenaud good catch! I'll add tag filtering.

pkazmierczak:
> Is there ever a situation where you'd have 2 registries with the same name but different `@version` suffixes on them?
> 
> Since we show the version suffix under `REGISTRY NAME` in the list output, I thought I'd try including it in my `--registry` flag, but it only cares about the registry name proper, not w/ version.

I added `--ref` flag in 171122a, so you can pass a ref now, too. I opted for having a separate flag instead of parsing `regName@regRef` for consistency with other nomad pack commands.

shoenig (approved):
LGTM!

PR #379, opened 2023-07-10
cli: registry list upgrade warning

### LINKED ISSUE #370: Add a suggestion/warning when running "registry list" for the first time since behaviour change
If you run `nomad-pack registry list` after upgrading to the latest upcoming release, you'll be met with a list of packs that you, since #354 , would now expect to be seen with `nomad-pack list`, and with columns missing:

image

The user can fix this by running `nomad-pack registry delete default` (or whatever non-default registry they happened to have), but there's no indication that this is required when they first run the new `registry list`. This could be confusing for existing pack users.

Could we detect if `registry list` lists packs, not registries, as indicated by the above screenshot, and suggest to the user that they run a `registry delete ______` along a message that the command has changed? If this message gets shown conditionally when getting the above column-missing output, it seems like it would be unobtrusive but helpful.

pkazmierczak:
**Description**

`registry list` should show an upgrade message if it detects a user has an older format of registry in the cache.

New behavior:
```
~/pack $ nomad-pack registry list
  REGISTRY NAME |                REF                | LOCAL REF | REGISTRY URL
----------------+-----------------------------------+-----------+---------------
  default       | alertmanager@latest               |           |
  default       | aws_ebs_csi@latest                |           |
  default       | aws_efs_csi@latest                |           |

It appears that you have a cache created before Nomad-Pack 0.1 (hence the
missing values in LOCAL REF and REGISTRY URL columns). We recommend deleting
your current cache and re-adding your registries to take full advantage of new
`nomad-pack registry list` and `nomad-pack list` commands behavior.
```

Resolves #370

### DIFF
--- a/internal/cli/registry_list.go
+++ b/internal/cli/registry_list.go
@@ -53,6 +53,15 @@ func (c *RegistryListCommand) Run(args []string) int {
 			table.Rows = append(table.Rows, tableRow)
 		}
 		c.ui.Table(table)
+
+		// TODO: This message is to make upgrading from tech preview versions
+		// to 0.1 easier. Remove it at 0.2 release.
+		if table.Rows[0][2].Value == "" {
+			c.ui.Warning("It appears that you have a cache created before Nomad-Pack 0.1 (hence the\n" +
+				"missing values in LOCAL REF and REGISTRY URL columns). We recommend deleting\n" +
+				"your current cache and re-adding your registries to take full advantage of new\n" +
+				"`nomad-pack registry list` and `nomad-pack list` commands behavior.")
+		}
 	} else {
 		c.ui.Output("No registries present in the cache.")
 	}


internal/cli/registry_list.go
```diff
@@ -53,6 +53,15 @@ func (c *RegistryListCommand) Run(args []string) int {
...
 		}
 		c.ui.Table(table)
+
+		// TODO: This message is to make upgrading from tech preview versions
+		// to 0.1 easier. Remove it at 0.2 release.
+		if table.Rows[0][2].Value == "" {
+			c.ui.Warning("It appears that you have a cache created before Nomad-Pack 0.1 (hence the\n" +
+				"missing values in LOCAL REF and REGISTRY URL columns). We recommend deleting\n" +
+				"your current cache and re-adding your registries to take full advantage of new\n" +
+				"`nomad-pack registry list` and `nomad-pack list` commands behavior.")
```
philrenaud, line 63: this was our chance to sneak `behaviour` instead of `behavior` in there while the americans weren't looking, Piotr!

philrenaud (approved):

internal/cli/registry_list.go
```diff
@@ -53,6 +53,15 @@ func (c *RegistryListCommand) Run(args []string) int {
 			table.Rows = append(table.Rows, tableRow)
 		}
 		c.ui.Table(table)
+
+		// TODO: This message is to make upgrading from tech preview versions
+		// to 0.1 easier. Remove it at 0.2 release.
+		if table.Rows[0][2].Value == "" {
+			c.ui.Warning("It appears that you have a cache created before Nomad-Pack 0.1 (hence the\n" +
+				"missing values in LOCAL REF and REGISTRY URL columns). We recommend deleting\n" +
+				"your current cache and re-adding your registries to take full advantage of new\n" +
```
shoenig, line 62: how does one delete your current cache?

pkazmierczak: `nomad-pack registry delete $registryName`

internal/cli/registry_list.go
pkazmierczak: I really wanted to do this, but as I typed `behaviour` I saw a bald eagle flying past and I heard Mariah Carey's voice singing "oh say can you see," and I changed it to `behavior`.

shoenig (approved):

PR #380, opened 2023-07-11
cli: add an overwrite flag for generate pack cmd

### LINKED ISSUE #376: nomad-pack generate overwrites an existing pack without warning
Would be nice to let me know that the pack I'm about to generate already exists.

Steps to reproduce:
- Create a pack (via `generate` or manually) named pack1
- make edits to `metadata.hcl`, `templates/pack1.nomad.tpl`, etc.
- `nomad-pack generate pack pack1`
results in `Creating "pack1" Pack in "."...` without indication that a pack already exists there.

pkazmierczak:
**Description**

This PR adds an `--overwrite` flag for `generate pack` command (defaults to `false`). This way we avoid accidentally overwriting files. 

Resolves #376 

**Reminders**


philrenaud (approved):
Works like a charm!

PR #381, opened 2023-07-11
makefile: install dev binary by default

pkazmierczak:
Could we make the `dev` target install the binary the way Nomad does it? Would make it a more consistent developer experience between the two.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -69,16 +69,10 @@ hclfmt: ## Format HCL files with hclfmt
 
 .PHONY: dev
 dev:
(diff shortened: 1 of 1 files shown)


angrycub:
Seems like opting in to the automatic install is least surprising and only needs to be done once for developers wishing to immediately run their dev builds. Stepping on the version that a user installs with `go install` feels bad unless they are expecting it.

IMO, this pattern would be good for Nomad to adopt rather than making the automatic dev build installation opt-out in the Pack makefile.

PR #384, opened 2023-07-17
[chore] sdk package should be independent

pkazmierczak:
`sdk` should be independent from nomad-pack internal packages. Eliminates a warning from `make check`, too.


jrasell (approved):
LGTM, thanks!

PR #385, opened 2023-07-17
fix: add nomad as test dependency to fix failing CLI tests

pkazmierczak:
Since [upgrading Nomad API to 1.5](https://github.com/hashicorp/nomad-pack/pull/383), we require `nomad` binary present to run CLI tests. This is due to the fact that Nomad 1.5+ [sandboxes go-getter](https://github.com/hashicorp/nomad/pull/15328) and thus using the taskrunner in tests requires an actual Nomad binary.

### DIFF
--- a/.github/workflows/go-tests.yml
+++ b/.github/workflows/go-tests.yml
@@ -40,10 +40,10 @@ jobs:
       # install nomad
       - name: Install nomad
         run : |
-          apt -y install wget gpg coreutils
+          sudo apt -y install wget gpg coreutils
           wget -O- https://apt.releases.hashicorp.com/gpg | gpg --dearmor -o /usr/share/keyrings/hashicorp-archive-keyring.gpg
-          echo "deb [signed-by=/usr/share/keyrings/hashicorp-archive-keyring.gpg] https://apt.releases.hashicorp.com $(lsb_release -cs) main" | tee /etc/apt/sources.list.d/hashicorp.list
-          apt update && apt -y install nomad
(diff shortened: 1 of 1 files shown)


.github/workflows/go-tests.yml
```diff
@@ -37,6 +37,15 @@ jobs:
           go install gotest.tools/gotestsum@latest
           make test-certs
 
+      # install nomad
+      - name: Install nomad
```
jrasell, line 41:
```suggestion
      - name: Install Nomad
```

.github/workflows/go-tests.yml
```diff
@@ -37,6 +37,15 @@ jobs:
...
+        run : |
+          sudo apt -y install wget gpg coreutils
+          wget -O- https://apt.releases.hashicorp.com/gpg | gpg --dearmor -o /usr/share/keyrings/hashicorp-archive-keyring.gpg
+          echo "deb [signed-by=/usr/share/keyrings/hashicorp-archive-keyring.gpg] https://apt.releases.hashicorp.com $(lsb_release -cs) main" | sudo tee /etc/apt/sources.list.d/hashicorp.list
+          sudo apt update && sudo apt -y install nomad
```
jrasell, line 46: Do we want to pin this to a version, or does that not really add anything nor matter?

jrasell (approved):
LGTM with a non-blocking question.

.github/workflows/go-tests.yml
pkazmierczak: I thought about this, but i actually want latest stable. Because if our tests start failing on latest stable, I wanna know.

PR #386, opened 2023-07-18
cli: generate varfile should write to file by default

### LINKED ISSUE #378: Consider changing `generate var-file` to `render var-file`
Caveat: I am new enough to `nomad-pack` that this may be a convention that oughtn't be changed, so please feel free to ignore this request.

Since the default behaviour of `nomad-pack generate var-file` is console output, and only by using the `--to-file` flag would you generate a new file, should we consider moving its command space to `nomad-pack render var-file` ?

The current behaviour seems notably different than `nomad-pack generate pack`, which does generate local file and directory changes.

pkazmierczak:
**Description**

In order to be consistent with other `generate` commands, `generate var-file` should default to writing to file, and optionally print to stdout. 

Resolves #378

### DIFF
--- a/internal/cli/generate_varfile.go
+++ b/internal/cli/generate_varfile.go
@@ -25,12 +25,14 @@ type generateVarFileCommand struct {
 	*baseCommand
 	packConfig *cache.PackConfig
 
-	// renderTo is the path to write rendered var override files to in
-	// addition to standard output.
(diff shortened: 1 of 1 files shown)


jrasell (approved):
LGTM!

philrenaud (approved):

PR #392, opened 2023-07-28
[chore] fix nightly-release.yml

angrycub:
The github-script action v6.4.1 introduced breaking changes from v3, and the script itself was not updated to reflect them causing nightly builds to fail.

This commit also downscopes the generated github token.

I have manually triggered a run [Nightly Release #425](https://github.com/hashicorp/nomad-pack/actions/runs/5694867288) using this version of the action and it was successful.


gulducat (approved):
:100:

PR #393, opened 2023-07-28
[bug] Handle nil `Subject` in `Diags` wrapper

angrycub:
**Description**
It is possible to receive an `hcl.Diagnostic` that does not have `Subject` set, which causes the existing code to panic. Added a nil check around it


gulducat (approved):
:relieved: (this emoji is not panicking)

PR #397, opened 2023-08-03
[chore] Bump makefile installed golangci-lint to v1.51.2

angrycub:
**Description**
Updating the version installed by the makefile to match the version used in the GitHub action.


lgfa29 (approved):
Ah interesting. I would expect GHA to call this make target 😄

PR #399, opened 2023-08-04
[deps] nomad co-dependency updates

angrycub:
- replaces hashicorp/hcl with the same version Nomad does.
- switches hashicorp/hcl2 to the patched version that Nomad runs.


gulducat (approved):
sure would be neat if this weren't required, but c'est la vie.

PR #400, opened 2023-08-04
[chore] bump golang.org/x/net v0.12.0 -> v0.13.0

angrycub:
Bumping golang.org/x/net v0.12.0 -> v0.13.0 to prevent the security scanner from complaining.

```
OSV reported vulnerability GO-2023-1988 in golang.org/x/net@v0.12.0
```

This library is not used directly in Pack, but is included indirectly as a test dependency through Pack's use of Nomad's testagent package.


gulducat (approved):
:closed_lock_with_key:

PR #401, opened 2023-08-04
[chore] Update `Test` action to use environment files

angrycub:
**Description**
GHA is [transitioning away](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) from set-output in favor of [setting outputs via "environment files"](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter). This PR makes the recommended changes to the `Test` action.


angrycub:
Closing in favor of #402

PR #402, opened 2023-08-04
[chore] use hashicorp/setup-golang action

gulducat:
**Description**

[hashicorp/setup-golang](https://github.com/hashicorp/setup-golang) uses `.go-version` file and does caching by default, so we can drop some extra cruft here.  Since it's in our GitHub org, we also don't need to pin an action SHA either, which is kinda nice.

I'm avoiding `build.yml` just for personal bandwidth reasons, but maybe this can serve as inspiration for anyone wishing to compress some of that too. :heart:


angrycub (approved):
LGTM :shipit:

PR #403, opened 2023-08-08
[feature] Nested Pack updates

angrycub:
# Improved Nested Pack Support

## Abstract
To better support nested pack dependencies, this PR takes a different approach to the way in which dependency packs are stored within one another and how the template context is organized. It does, however, retain the ability to process un-updated packs using a "--parser-v1" flag.

## New variable naming scheme

When supplying variables to the pack, the names must be provided using fully-qualified dotted references. For example, you would use the following command to specify the value of variable "dcs" on a file-system based, root pack named "app" using the `--var` flag

```
nomad-pack render --var app.dcs='["dc1"] .
```

### Environment Variables

For envvars this is more tricky, dots are problematic, but almost any substitution you could make would yield an ambiguous name to parse. For now, variables can be read from the environment as `NOMAD_PACK_VAR_app.dcs` and have to be provided by a non-shell wrapper like `env` or custom tooling. There will be future work to try and iron out this particular wrinkle.

### Implementation notes

There is a lot of magic that happens in the variable override file parsing to allow for dotted left-hand side variables in the HCL. It leverages a cheat based on how HCL deals with map keys versus attributes and wraps the filedata in map syntax. Any produced hcl.Diagnostics based on that data have to be touched up before being presented to the end user. This is "clever" code, where clever in this case is probably a dirty word.

## Supporting the new and current syntax

One of the other feature(esque) components of this PR is to provide some time for adopters to start migrating their packs to the new syntax by continuing to support the current syntax. I've been told to rip the bandaid off, and I can appreciate that from a code complexity standpoint. However, it made me feel better about unleashing a new template context shape and supportive functions knowing that I was less likely to leave someone fully broken if we could support the current syntax.

### template_context.go
Currently, there is a non-trivial amount of entanglement in the template_context.go that I don't love, and the parser/renderer abstraction is super leaky there. But since it's current state is transitory and we wanted to move this along, it's grody but functioning.

### funcMaps
In supporting both versions of pack syntax, I wanted to try to give Pack authors and consumers as much feedback for when (not if) something gets crossed up between the syntaxes. When using the v1 parser, the funcmap is built with functions having the same name as the v2 parser that return meaningful runtime errors to the end user.

### PackTemplateError
This is an attempt to help users who run a pack having v1 syntax using the v2 parser. Because the contexts are completely differently shaped, the most likely error that will happen is a nil pointer exception that leaks the PackContextable detail during template rendering. Go's text/template errors are certainly not designed to be end-user facing and PackContextable  is an internal implementation detail, so the plan here is to make more user friendly errors starting with template execution errors. There are opportunities to grow this out over time if it's desirable.

## Refactoring

This is the part of this PR that profoundly bloats it. 

### Typed map keys

While trying to reorganize the template context, it became nearly impossible to track exactly what was going on in the `map[string]any` and map[string]map[string]any` that I wanted to come back and move to typed map keys. This change was _comically_ sprawly.

### Test fixture rewrites and reorganization

Since there are now two syntaxes to parse, I brought back the tests for the earlier syntax. I reorganized the fixtures so that when the time comes, the  v1 fixtures should be easily removed. 

Several tests were printing out the test case name in their error rather than taking advantage of `t.Run` to build subtests, so many test cases were updated to switch to subtests.

### Rehoming some errors and HCL diags

There were some opportunities to move some of the errors and hcl.Diagnostics to the common errors package.


jrasell (commented):
This is amazing work @angrycub!

There were a few empty files and wondered if we can get rid of these?

In a number of places we use `fmt.Errorf` but don't format the error, so it would be nice to use `errors.New` instead.

I think some files are missing the license header.

It might be nice to have smaller unit tests alongside the bigger ones. I noticed a few functions did not have direct tests, which can sometimes be useful.

`_self` is used a bit and while I noted it in the code, I thought it was worth asking whether this should be a "top level" const/type somewhere?

The work to support v1/v2 parsers is huge and I think the rationale you wrote makes sense, however, I do wonder how this compares to the maintainability overhead moving forward. If this is a stop-gap while packs move over to the new style, then I think we should call this out in a number of places, and plan to remove it. Alternatively, could old style pack users just use the older releases, so it creates a clear cut?

internal/pkg/errors/template_errors.go
```diff
@@ -0,0 +1,215 @@
...
+
+	var execErr template.ExecError
+	if errors.As(err, &execErr) {
+		out.parseExecError(execErr)
+	}
+
+	return out
+}
+
+func (p PackTemplateError) ToWrappedUIContext() *WrappedUIContext {
```
pkazmierczak, line 45: vide James' comment above: make methods of `PackTemplateError` consistent, i.e., they should all be pointer receivers.

internal/pkg/renderer/renderer.go
```diff
@@ -87,7 +113,7 @@ func (r *Renderer) Render(p *pack.Pack, variables map[string]any) (*Rendered, er
 	// Generate our output structure.
 	rendered := &Rendered{
 		parentRenders:    make(map[string]string),
-		dependentRenders: make(map[string]string),
+		dependentRenders: make(map[string]string), // TODO: Are these "dependencyRenders"?
```
pkazmierczak, line 116: yeah I think they are

internal/pkg/varfile/varfile_test.go
```diff
@@ -0,0 +1,215 @@
...
+				for i, o := range diags {
+					e := tc.exp.diags[i]
+					must.Eq(t, e, o)
+				}
+			}
+			// There should always be a single element map for these tests having the
+			// filename as the key
+			switch tc.exp.oLen {
+			case 0:
+				must.MapLen[variables.Overrides](t, 0, om, must.Sprintf("map contents: %v", spew.Sdump(om)))
```
pkazmierczak, line 110: I think `spew`ing the contents here is maybe a bit much, we could just have a small helper function that would output map contents. I think it's just best to avoid pulling spew as a dependency because it's a big one, but then again I think templates pull it as a dependency anyway, so I don't think strongly about this.

internal/pkg/varfile/varfile_test.go
```diff
@@ -0,0 +1,215 @@
...
+			d2 := DecodeResult{
+				Overrides: variables.Overrides{},
+			}
+			dr.Merge(d2)
+			must.False(t, dr.Diags.HasErrors())
+			must.Len[*variables.Override](t, 2, dr.Overrides["p2"])
+			must.MapNotContainsKey(t, d2.Overrides, "p1")
+		})
+
+		//TODO: Investigate this broken test.
```
pkazmierczak, line 198: Is this something we should still investigate?

internal/pkg/variable/parser.go
```diff
@@ -1,268 +1,20 @@
...
-	// packs that are being parsed. The first map is keyed by the pack name,
-	// the second is by the variable name.
-	rootVars map[string]map[string]*Variable
-
-	// envOverrideVars, fileOverrideVars, cliOverrideVars are the override
-	// variables. The maps are keyed by the pack name they are associated to.
-	envOverrideVars  map[string][]*Variable
-	fileOverrideVars map[string][]*Variable
-	cliOverrideVars  map[string][]*Variable
+type Parser interface {
```
pkazmierczak, line 10: I'm not sure about the usefulness of this interface. In the code we only ever use it once: in the `NewParser` constructor as its return type, but then I wonder: as a caller of a new parser constructor, wouldn't I want to be able to assert the type of the parser (v1/v2)? Perhaps there's something I'm missing here but I feel this interface is just superfluous.

internal/pkg/variable/parser/parsed_variables.go
```diff
@@ -0,0 +1,192 @@
...
+	for k, vs := range in {
+		out[PackID(k)] = make(map[VariableID]*Variable, len(vs))
+		for vk, v := range vs {
+			out[PackID(k)][VariableID(vk)] = v
+		}
+	}
+	return out
+}
+
+// NOTE: Beyond here, things get weird.
```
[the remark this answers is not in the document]
pkazmierczak: From discussions with @angrycub I gather it would just be very very complicated explaining this in a comment :) Still, probably worth it.

internal/pkg/variable/parser/parser.go
```diff
@@ -0,0 +1,20 @@
...
+// SPDX-License-Identifier: MPL-2.0
+
+package parser
+
+import (
+	"github.com/hashicorp/hcl/v2"
+	"github.com/hashicorp/nomad-pack/internal/pkg/variable/parser/config"
+)
+
+type Parser interface {
```
pkazmierczak, line 11: sooo we have `internal/pkg/variable/parser.go:Parser` and `internal/pkg/variable/parser/parser.go:Parser` interfaces? 🤔 
Also my comment from the previous `Parser` interface definition above still applies here.

internal/pkg/variable/parser/template_context.go
```diff
@@ -0,0 +1,316 @@
...
+// │       │ │       │ │       │ ┌────────────────────┐ │ │ │
+// │       │ │       │ │ _self │ │PackData            │ │ │ │
+// │       │ │       │ │       │ ├────────────────────┤ │ │ │
+// │       │ │       │ │       │ │Pack *pack.Pack     │ │ │ │
+// │       │ │       │ │       │ │meta map[string]any │ │ │ │
+// │       │ │       │ │       │ │vars map[string]any │ │ │ │
+// │       │ │       │ │       │ └────────────────────┘ │ │ │
+// │       │ │       │ └───────┴────────────────────────┘ │ │
+// │       │ └───────┴────────────────────────────────────┘ │
+// └───────┴────────────────────────────────────────────────┘
```
pkazmierczak, line 83: 🙏 thanks for this diagram, very nice.

sdk/pack/variables/map_test.go
```diff
@@ -0,0 +1,17 @@
+package variables
+
+import (
+	"testing"
+
+	"github.com/shoenig/test/must"
+)
+
+func TestIDKeyedVarMap_Variables(t *testing.T) {
+	// TODO: needs to be implemented
```
pkazmierczak, line 10: Would be nice to have those indeed. I'm fine with these tests coming in a separate PR though.

sdk/pack/variables/variables.go
```diff
@@ -0,0 +1,156 @@
...
+		cv.Default == ov.Default &&
+		cv.hasDefault == ov.hasDefault &&
+		cv.Type == ov.Type &&
+		cv.hasType == ov.hasType &&
+		cv.Value == ov.Value
+
+	return eq
+}
+
+func (v Variable) AsOverrideString(pID pack.ID) string {
```
pkazmierczak, line 77: value receiver as opposed to all other methods being pointer receivers (vide James' comment above)

pkazmierczak (approved):
Monumental work, @angrycub! I left a few comments that I'd like to see addressed before merging, but I am approving, I think in its current shape the PR looks good to be merged. 

Once again, the scope of improvements is truly amazing, congrats! 🎉

sdk/pack/variables/variables.go
```diff
@@ -0,0 +1,160 @@
...
+
+type Variables []*Variable
+
+func (v Variables) Len() int           { return len(v) }
+func (v Variables) Less(i, j int) bool { return v[i].Name < v[j].Name }
+func (v Variables) Swap(i, j int)      { v[i], v[j] = v[j], v[i] }
+
+// Variable encapsulates a single variable as defined within a block according
+// to variableFileSchema and variableBlockSchema.
+type Variable struct {
```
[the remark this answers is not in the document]
angrycub: I think `Equal` was the primary offender. Updated in 0e06d6973eb1e22250e9fd842f233b591af0322f

PR #404, opened 2023-08-09
[tests] Update cache tests

angrycub:
**Description**
Found that adding a pack to the fixtures/test_registry broke the cache tests. This PR adds test helpers to more accurately assess what is in the test cache directories.

**Commits**
- Adding a pack to the registry breaks tests
- Remove shortening from refs
- Refactor `testPackCount` into `listAllTestPacks` and helper funcs


pkazmierczak (approved):
LGTM!

PR #413, opened 2023-08-22
[ui] Reshorten refs for output

angrycub:
**Description**
In #404, I made a drive-by change to use the full version of the localRef—typically a SHA1 hash. This caused the output of the `nomad-pack list` command to become extremely wide. This PR adds a helper function to reduce them back to 8 characters when printed.


pkazmierczak (approved):
nice!

PR #420, opened 2023-09-08
resolves #320 fix error when formatting HCL2 templates

### LINKED ISSUE #320: nomad hcl functions can't be used in a nomad pack
Hello :)

Since two weeks, I have a problem with the rendering of my packs with the nightly build.
I get the following error :
```
Error: failed to format the template ... , At : Unknown token: 11:25 IDENT <function_name>
```

It seems to be caused by this PR (https://github.com/hashicorp/nomad-pack/issues/301) that added formatting to nomad job files.
And it seems to fail when we use any nomad hcl functions: https://developer.hashicorp.com/nomad/docs/job-specification/hcl2/functions

When we add --no-format, everything works.

The following job definition (in a pack) can't be rendered:
```
job "test" {
    group "test" {
        task "test" {
            driver = "docker"

            config {
                image = "busybox"
            }

            env {
                VALUE=max(10,20)
            }
        }
    }
}
```

slonopotamus:
The issue was that nomad-pack used HCL1 formatter.

### DIFF
--- a/internal/pkg/renderer/renderer.go
+++ b/internal/pkg/renderer/renderer.go
@@ -9,7 +9,7 @@ import (
 	"strings"
 	"text/template"
 
-	"github.com/hashicorp/hcl/hcl/printer"
+	"github.com/hashicorp/hcl/v2/hclwrite"
 	"github.com/hashicorp/nomad/api"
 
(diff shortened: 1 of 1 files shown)


pkazmierczak (approved):
LGTM! Thanks for the contribution @slonopotamus!

PR #423, opened 2023-09-13
fix: parse userinfo provided in nomad address

stind:
**Description**

I was trying to use `nomad-pack` with address that includes basic auth, like `http://user:pass@host`, and received `nil pointer dereference` error. 

On top of that, I realized that the address is not parsed correctly when it includes both `scheme://` and userinfo: `scheme://user:pass@addr` would be parsed as `user = "scheme"`, `pass = "//user:pass"`, `addr = "addr"`

The current workaround is to set `NOMAD_HTTP_AUTH` env variable with `user:pass`.

**Reminders**


stind:
> This is kinda... Eh. Maybe just `url.Parse`?

more than happy to do that 👍 I'll make the change

angrycub:
@stind Thanks for the PR. Typically the Nomad API client is configured using the [`DefaultConfig`](https://github.com/hashicorp/nomad/blob/main/api/api.go#L292) function of the API package.
Since the Nomad CLI uses `NOMAD_HTTP_AUTH` to hold the basic authentication information, we do the same for consistency's sake.

However, I think it might be nice to be able to consume all of this from the NOMAD_ADDR itself, so I'd like to talk to some folks here and see if this is something that we can bring into the Nomad API package itself.

PR #426, opened 2023-09-22
[f-nested-deps] Remove requirement to have root pack name in overrides

angrycub:
Continues on the work in #403. Removes the requirement to specify the name of the root pack in the override variable's path.


internal/pkg/varfile/varfile.go
```diff
@@ -13,13 +13,15 @@ import (
 	"github.com/hashicorp/nomad-pack/sdk/pack/variables"
 )
 
-func DecodeVariableOverrides(files []*pack.File) DecodeResult {
+// TODO: this is only used in a test - remove?
```
pkazmierczak, line 16: Remove.

internal/pkg/varfile/varfile.go
```diff
@@ -217,9 +221,30 @@ func decode(filename string, src []byte, ctx *hcl.EvalContext, target *variables
 		oRange := hcl.RangeBetween(kv.Key.Range(), kv.Value.Range())
 		fixupRange(&oRange)
 
+		var path pack.ID
+		var name variables.ID
+
+		// NOTE: This implementation assumes a single element variable value,
```
pkazmierczak, line 227: I think that's fine for now but maybe make an issue so that we could revisit later?

pkazmierczak (approved):
LGTM!

internal/pkg/varfile/varfile.go
angrycub: Created issue #435 to capture this

internal/pkg/varfile/varfile.go
angrycub: Moved into the test package

PR #434, opened 2023-10-02
[copywrite] Ignore creator fixtures

angrycub:
**Description**

Add a `copywrite` configuration that ignores the `internal/creator/templates` files, since they are stub pages intended for end-user content.  Once we merge this, we need to have the bot rerun PR #433


pkazmierczak (approved):
Good catch, LGTM

PR #440, opened 2023-10-06
[generator] Fix generated job spec bug

angrycub:
**Description**

While updating the creator package's templates, I missed one of the variable references so as it stands the sample job will never be able to be Consul-enabled. 

**Reminders**


pkazmierczak (approved):
Nice catch, LGTM!

PR #441, opened 2023-10-09
[cli] Update version output to align with Nomad

angrycub:
**Description**
Update to the version output to match the output given from the Nomad output command

### DIFF
--- a/internal/pkg/version/version.go
+++ b/internal/pkg/version/version.go
@@ -4,65 +4,110 @@
 package version
 
 import (
+	"bytes"
 	"fmt"
-	"strings"
+	"time"
 )
 
 var (
+	// BuildDate is the time of the git commit used to build the program,
+	// in RFC3339 format. It is filled in by the compiler via makefile.
+	BuildDate string
+
 	// GitCommit and GitDescribe are filled by the compiler using ldflags to
 	// provide useful Git information.
 	GitCommit   string
 	GitDescribe string
 
 	// Version is the semantic version number describing the current state of
-	// NOM.
+	// nomad-pack.
 	Version = "0.0.1"
 
 	// Prerelease designates whether the current version is within a prerelease
 	// phase. Typically, this will be "dev" to signify a development cycle or a
 	// release candidate phase such as alpha, beta.1, rc.1, or such.
-	Prerelease = "techpreview.4"
+	VersionPrerelease = "techpreview.4"
 
 	// Metadata allows us to provide additional metadata information to the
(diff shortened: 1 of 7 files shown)


angrycub:
Since @gulducat did the Nomad implementation, adding as a reviewer

internal/pkg/version/version.go
```diff
@@ -4,65 +4,110 @@
...
-			// already.
-			version += fmt.Sprintf("-%s", release)
-		}
-
-		if Metadata != "" {
-			version += fmt.Sprintf("+%s", Metadata)
-		}
+		ver = GitDescribe
+	}
+	if GitDescribe == "" && rel == "" && VersionPrerelease != "" {
```
gulducat, line 62: I don't really want to open up a can of `build.yml` worms, because I don't think this _actually_ matters, but `GitDescribe` doesn't seem useful?  [in build.yml](https://github.com/hashicorp/nomad-pack/blob/cbf90695c94ad108ad59c6eed3e23c9f6e91468e/.github/workflows/build.yml#L69) it just gets set to a previous job's `product-version` output, which is the same (via `make version` => `scripts/version.sh`) as `VersionNumber()` in this very file?  seems very circuitous...

but maybe we just ignore it, because there seems like an awful lot of cruft in build.yml, so it could be full of refactorable rabbit holes, and it does work fine enough. ¯\\\_(ツ)\_/¯

internal/pkg/version/version.go
```diff
@@ -4,65 +4,110 @@
...
+		version = fmt.Sprintf("%s+%s", version, c.VersionMetadata)
+	}
+
+	return version
+}
+
+func (c *VersionInfo) FullVersionNumber(rev bool) string {
+	var versionString bytes.Buffer
+
+	fmt.Fprintf(&versionString, "Nomad v%s", c.Version)
```
gulducat, line 95: dropped a "Pack" here

```suggestion
	fmt.Fprintf(&versionString, "Nomad Pack v%s", c.Version)
```

looks to be the cause of [this test failure](https://github.com/hashicorp/nomad-pack/actions/runs/6458991151/job/17533963690?pr=441#step:3:369) too.

gulducat: that ^ will re-break "fix tests" b9ffbe184f83dc15b4608764ade6801089d01d5c I think, but I figure we do want "Nomad Pack" in the output here:

```
$ ./bin/nomad-pack version
Nomad v0.0.1-techpreview.4
BuildDate ...etc...
```

PR #444, opened 2023-10-20
Use shallow cloning to add pack registries

attachmentgenie:
We use a mono repo to build out all of our platforms, as a down side of this our git history and .git directory are huge (think 10x the actual code size)which causes git clones to be quite slow.

This PR switches the git clone behaviour of `nomad-pack registry add` to use shallow clone instead. In our particular case it drops the length of the process from ~9s to 0.5s.

**Reminders**

### DIFF
--- a/internal/pkg/cache/add.go
+++ b/internal/pkg/cache/add.go
@@ -154,6 +154,9 @@ func (c *Cache) cloneRemoteGitRegistry(opts *AddOpts) (string, error) {
 	// If ref is set, add query string variable
 	if !opts.IsLatest() {
 		url = fmt.Sprintf("%s?ref=%s", url, opts.Ref)
+	} else {
+		// Attempt to shallow clone the constructed url
+		url = fmt.Sprintf("%s?depth=1", url)
 	}
 
 	logger.Debug(fmt.Sprintf("go-getter URL is %s", url))
@@ -163,8 +166,7 @@ func (c *Cache) cloneRemoteGitRegistry(opts *AddOpts) (string, error) {
 	if opts.PackName != "" {
 		clonePath = path.Join(clonePath, "packs", opts.PackName)
 	}
-	// Attempt to shallow clone the constructed url
-	if err := gg.Get(clonePath, fmt.Sprintf("git::%s?depth=1", url)); err != nil {
+	if err := gg.Get(clonePath, fmt.Sprintf("git::%s", url)); err != nil {
 		logger.ErrorWithContext(err, "could not install registry", c.ErrorContext.GetAll()...)
 		return "n/a", err
 	}


internal/pkg/cache/add.go
```diff
@@ -154,6 +154,9 @@ func (c *Cache) cloneRemoteGitRegistry(opts *AddOpts) (string, error) {
 	// If ref is set, add query string variable
 	if !opts.IsLatest() {
 		url = fmt.Sprintf("%s?ref=%s", url, opts.Ref)
+	} else {
+		// Attempt to shallow clone the constructed url
+		url = fmt.Sprintf("%s?depth=1", url)
```
pkazmierczak, line 159: I'd rather have the shallow clone be a a flag to the add command than handling it this way.

pkazmierczak (changes requested):
Thanks for the PR @attachmentgenie! This is good thinking, in many scenarios (big repo, using pack in CI) a shallow clone makes much more sense than a full clone. However, shallow clones affect what we can do with the repository, and that's why I'd prefer shallow-cloning to be a flag. 

For example, if we were to support updating registries, we'd have to introduce special treatment for shallow-cloned repositories, and their updates would take much longer. Secondly, users might want to switch branches of their registries (granted, a niche use-case, but possible with current implementation and defaulting to shallow-clones would  break it). 

I don't have a strong opinion on whether this should be an opt-in or an opt-out flag, perhaps @angrycub has an opinion? But I think it should be a flag nonetheless.

attachmentgenie:
I debated with myself about feature flagging it but ultimately (lazily) decided not to do it because as long as , [this bit of code](https://github.com/hashicorp/nomad-pack/blob/db9f7f8c84da83f267d9f6c4d1d103ba7c395a7c/internal/pkg/cache/add.go#L77-L81) is in place, there are IMHO no benefits to keeping any history more that the commit your are looking for since nomad-pack cleans that up for you automatically after adding the registry

But i am more than happy to add the feature flag code if you are considering removing that code in the (near) future

pkazmierczak:
> as long as , [this bit of code](https://github.com/hashicorp/nomad-pack/blob/db9f7f8c84da83f267d9f6c4d1d103ba7c395a7c/internal/pkg/cache/add.go#L77-L81) is in place, there are IMHO no benefits to keeping any history more that the commit your are looking for since nomad-pack cleans that up for you automatically after adding the registry

Right! The team has been thinking about adding an [update feature](https://github.com/hashicorp/nomad-pack/issues/446) though, and I just thought fetching would be a natural way to do this. 

The more I think about shallow clones, the more I think this could just be an opt-out feature, as in: we'd default to a shallow clone.

attachmentgenie:
> The more I think about shallow clones, the more I think this could just be an opt-out feature, as in: we'd default to a shallow clone.

alright, will update to do just that

pkazmierczak (approved):
I've reconsidered, you're right. Shallow-clones are the way to go. Vide https://github.com/hashicorp/nomad-pack/pull/447#pullrequestreview-1697680506

PR #445, opened 2023-10-23
[cli] Fix `generate registry` command

angrycub:
**Description**

The argument for the `generate registry` command is improperly read causing the command to emit the generated files into the current folder rather than a folder having the name of the registry.

**Reminders**


pkazmierczak (approved):
LGTM!

PR #447, opened 2023-10-24
feat: allow dependencies to be pinned to a git ref

attachmentgenie:
**Description**

We would prefer our dependencies to be pinned so that rerunning `nomad-pack deps vendor` would always yield the same result.

Also a small fix for the template, the current proposed source string errors out. This also includes the shallow cloning patch as applied in #444 for registries as a whole

**Reminders**


pkazmierczak (approved):
Hey @attachmentgenie, this is my feedback for this PR and for https://github.com/hashicorp/nomad-pack/pull/444: let's do shallow cloning like you suggested. After discussing this with @angrycub we agree that shallow clones are the way to go. There's just no need for extra git info in our caches, and in general we like the direction of migrating away from git-centric registries and caches for pack. So: LGTM!

PR #449, opened 2023-10-24
[bug] Fix nested pack map key creation error

### LINKED ISSUE #448: Error setting a dependency variable value using the `--var` flag
Currently Nomad pack will error when a user attempts to set the value for a dependency's variable using the command line as follows,
```
nomad-pack render fixtures/v2/test_registry/packs/deps_test_1 \
    --var job_name=A \
    --var child1.job_name=A.1 \
    --var child1.gc.job_name=A.1.a \
    --var child2.job_name=A.2 \
    --var child2.gc.job_name=A.2.a
```
returns an error from the `render` command.
```
! Failed To Process Pack

    Error:   There is no variable named "child1.gc.job_name". An override file can only override a variable that was
already declared in a primary configuration file.

... elided error output ...
```

Versus the equivalent var-file based output,
```
nomad-pack render fixtures/v2/test_registry/packs/deps_test_1 \
    --var-file fixtures/v2/override_files/deps_test_1/test1.hcl
```
which returns a successful render of the pack.
```
deps_test_1/deps_test.txt:

           job_name:A
    child1.job_name:A.1
 gc.child1.job_name:A.1.a
    child2.job_name:A.2
 gc.child2.job_name:A.2.a

... elided successful render output ...
```

angrycub:
**Description**
When building the fully-qualified pack name an additional dot was being included causing the flag-provided values to never match their corresponding pack 

Fixes #448

**Reminders**

- ~[ ] Add `CHANGELOG.md` entry~ Since this error was introduced and removed in the same version, I think we can omit the changelog line.

### DIFF
--- a/internal/cli/cli_test.go
+++ b/internal/cli/cli_test.go
@@ -491,7 +491,7 @@ func TestCLI_PackStatus_Fails(t *testing.T) {
 	})
 }
 
-func TestCLI_PackRender_MyAlias(t *testing.T) {
+func TestCLI_PackRender_RootVar(t *testing.T) {
 	t.Parallel()
 	// This test has to do some extra shenanigans because dependent pack template
 	// output is not guaranteed to be ordered. This requires that the test handle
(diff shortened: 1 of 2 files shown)


pkazmierczak (approved):
LGTM!

PR #451, opened 2023-10-24
Tweaks helpers and auto-generated pack

mikenomitch:
**Description**

Removes the mounts helper. There are already "complex" examples for helpers and this is probably too in the weeds and distracting for new users.

Also adds type to service helper since "nomad" is an option

Also adds dependency information into the metadata template

**Reminders**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,7 @@ bin/*
 *.dylib
 
 # binary when created in the project root with go build .
-./nomad-pack
+nomad-pack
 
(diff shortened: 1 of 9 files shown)


angrycub (approved):
Only a couple of concerns, but will be totally mergable w/o the comment changes; just need to get that test binary out... then :shipit:

angrycub (approved):
Good deal. 🎆 :shipit: 🎆

PR #456, opened 2023-10-26
[tests] Deflake the v1 exit code tests

angrycub:
**Description**
It appears that there is a timing related bug hiding in the original implementation. In this commit, I lifted the v2 version of the test after making some changes to dry it up a bit.


pkazmierczak (approved):
LGTM!

PR #462, opened 2023-10-27
[cli] Fix PackTemplateError

angrycub:
**Description**
When refactoring ParsedTemplates, it moved from variables to parser breaking the _extremely manual_ way in which the template errors were destructured into PackTemplateErrors. This restores the original behavior.

**Before**
Screenshot 2023-10-27 at 3 45 47 PM

**After**
Screenshot 2023-10-27 at 3 45 11 PM

- Fix PackTemplateError
- Switch to UIContext Constants
- Handle `my` alias; Add hint for --parser-v1

**Reminders**

### DIFF
--- a/internal/pkg/errors/template_errors.go
+++ b/internal/pkg/errors/template_errors.go
@@ -55,10 +55,10 @@ func ParseTemplateError(tplCtx parser.PackTemplateContext, err error) *PackTempl
 // display to the CLI
 func (p *PackTemplateError) ToWrappedUIContext() *WrappedUIContext {
 	errCtx := NewUIErrorContext()
-	errCtx.Add("Details: ", p.Details)
-	errCtx.Add("Filename: ", p.Filename)
-	errCtx.Add("Position: ", p.pos())
-	errCtx.Add("Suggestions: ", strings.Join(p.Suggestions, "; "))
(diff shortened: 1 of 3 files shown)


internal/pkg/errors/template_errors.go
```diff
@@ -167,32 +175,41 @@ func (p *PackTemplateError) enhanceNPE() {
...
-
 	}
 }
 
+func (p *PackTemplateError) isV2Error() bool {
+	return strings.HasSuffix(p.Err.Error(), "not implemented for nomad-pack's v1 syntax")
+}
+
+func (p *PackTemplateError) enhanceV2Error() {
+	p.Suggestions = []string{"Verify that the `--parser-v1` flag is not set when running this pack."}
```
tgross, line 200: Should this append to `p.Suggestions` so it doesn't clobber it?

```suggestion
	p.Suggestions = append(p.Suggestions, "Verify that the `--parser-v1` flag is not set when running this pack.")
```

tgross (approved):
LGTM!

internal/pkg/errors/template_errors.go
angrycub: It's safe right now, since the error conditions are mutually exclusive; but we could initialize the slice and switch to `append` if that's better. 🤔

PR #463, opened 2023-10-27
Promote Details and Suggestions to top level display

angrycub:
**Description**
Following the work in PR #462 , this PR consults the ErrorContext and if there are Details or Suggestions in it, extracts them to be shown immediately following the Error.

Screenshot 2023-10-27 at 2 55 19 PM

Screenshot 2023-10-27 at 3 26 07 PM

**Reminders**

### DIFF
--- a/terminal/noninteractive.go
+++ b/terminal/noninteractive.go
@@ -237,7 +237,7 @@ func (ui *nonInteractiveUI) ErrorWithContext(err error, sub string, ctx ...strin
 			key, rest, found := strings.Cut(item, ": ")
 
 			if !found {
-				wrapped := wordwrap.WrapString(rest, 78)
+				wrapped := wordwrap.WrapString(key, 78)
 				lines := strings.Split(wrapped, "\n")
 				for _, l := range lines {
 					ui.Error("  " + l)
--- a/terminal/noninteractive.go
+++ b/terminal/noninteractive.go
@@ -10,13 +10,16 @@ import (
 	"io"
 	"os"
 	"regexp"
+	"slices"
 	"strings"
 	"sync"
 	"text/tabwriter"
 
 	"github.com/fatih/color"
+	"github.com/mitchellh/go-wordwrap"
 	"github.com/olekukonko/tablewriter"
 
+	"github.com/hashicorp/nomad-pack/internal/pkg/errors"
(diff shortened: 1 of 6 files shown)


terminal/noninteractive.go
```diff
@@ -212,6 +215,51 @@ func (ui *nonInteractiveUI) Error(msg string) {
...
+		if oc, item, found := extractItem(ctx, key); found {
+			ctx = oc
+			if key == "" {
+				return
+			}
+
+			key, rest, found := strings.Cut(item, ": ")
+
+			if !found {
+				wrapped := wordwrap.WrapString(rest, 78)
```
tgross, line 240: https://pkg.go.dev/strings#Cut has returned `""` for `rest` if we're in this branch. Should this be?:

```suggestion
				wrapped := wordwrap.WrapString(key, 78)
```

terminal/noninteractive.go
```diff
@@ -212,6 +215,51 @@ func (ui *nonInteractiveUI) Error(msg string) {
...
+	var extractItem = func(ctx []string, key string) ([]string, string, bool) {
+		for i, v := range ctx {
+			if strings.HasPrefix(v, key) {
+				outStr := v
+				outCtx := slices.Delete(ctx, i, i+1)
+				return outCtx, outStr, true
+			}
+		}
+		return ctx, "", false
+	}
```
tgross, line 229: It seems like `extractItem` is shared in 3 places but the `testui` makes that hard to extract usefully. I wouldn't bother in this PR, but sometime in the future we might want to think about whether there's a set of common "ui helpers" we can factor out of these methods so we're not having to repeat them. (Ex. can noninteractive and testui share most of their code if they both take a writer that's configured by the caller.)

Definitely future work though!

tgross (approved):
LGTM, assuming the one item with the `rest` variable is either resolved or dismissed as me being wrong :grinning:

terminal/noninteractive.go
angrycub: Just noticed that I didn't update the glint UI implementation of `promote` to look like the non-interactive one. (It came later and that was when I heard @pkazmierczak's voice in my head, "Use strings.Cut()...".  Since it doesn't break anything, will make those the same in a follow-up PR.  Especially since I still want to come up with a hanging indent and wrapping story for the glint ui later anyway.

PR #467, opened 2023-11-01
[cli] Relax regex used for nil pointer enhancements

angrycub:
**Description**
When parsing a template execution error, the regex that extracted the location was too restrictive causing the decorated errors in any _-named helper template to not be rendered properly.

### DIFF
--- a/internal/pkg/errors/template_errors.go
+++ b/internal/pkg/errors/template_errors.go
@@ -220,11 +220,7 @@ func (p *PackTemplateError) fixupPackContextable() {
 		return
 	}
 
-	atRE, err := regexp.Compile(`^.*"` + p.Filename + `" at <(.+)>$`)
-	if err != nil {
-		// there's a janky filename that's breaking the regex.
-		return
-	}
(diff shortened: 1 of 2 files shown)


internal/pkg/errors/template_errors.go
```diff
@@ -220,11 +220,7 @@ func (p *PackTemplateError) fixupPackContextable() {
 		return
 	}
 
-	atRE, err := regexp.Compile(`^.*"` + p.Filename + `" at <(.+)>$`)
-	if err != nil {
-		// there's a janky filename that's breaking the regex.
-		return
-	}
+	atRE := regexp.MustCompile(`^executing .* at <(.+)>$`)
```
lgfa29, line 223: Since the regex seems static now you can declare it once at the top of the file so only need to compile once, but no biggie since it's a CLI tool

lgfa29 (approved):

internal/pkg/errors/template_errors.go
angrycub: Good call. Moved them up

PR #471, opened 2023-11-16
[jobspec2] Update HCL2 dependency

angrycub:
There is a bug in my port of Mahmood's fixes to HCL2. This bumps the
dependency to a version containing the fix.


angrycub:
Superseded by #514

PR #494, opened 2024-03-21
[docs] Add documentation for nomad-pack template functions

angrycub:
This PR starts chipping away at the Pack documentation tech debt by documenting the template helper functions that Pack provides to the template context.

### DIFF
--- a/docs/functions.md
+++ b/docs/functions.md
@@ -1,0 +1,567 @@
+# Nomad Pack functions
+
+The nomad-pack template renderer contains various helper functions used while rendering
+templates.
+
+Nomad Pack provides all the [Sprig template functions][Sprig] for text manipulation.
+
+Nomad Pack also provides additional functions for accessing Nomad, testing IP addresses, and
+template debugging.
+
+## Functions by topic
+
+- [Nomad API][topicNomadAPI]
+- [Debugging][topicDebugging]
+- [Helpers][topicHelpers]
+
+You can also view [the list of functions in alphabetical order][fnByAlpha]
+
+## Nomad API <a id="topicNomadAPI"></a>
+
+### Namespace functions
+
+#### `nomadNamespace` <a id="nomadNamespace"></a>
+
+The `nomadNamespace` returns the details for the namespace with the given name.
+
+**NOTE:** The `nomadNamespace` function will error on missing namespaces, which
+will prevent the template from rendering. Any part of the template that uses
+the `nomadNamespace` function should be using a namespace name provided by the
+`nomadNamespaces` function for additional safety.
+
+##### Parameters
+
+- 1: `string` - The target namespaces's name
+
+##### Returns
+
+- `error` or [`*api.Namespace`][] for the requested namespace.
+
+##### Example
+
+Get the name and description for each template accessible to the current user.
+
+```
+[[ range $ns := nomadNamespaces]]
+    [[ with nomadNamespace $ns ]]
+      [[printf "%s: %s\n" .Name .Description ]]
+    [[ end ]]
+[[ end ]]
+```
+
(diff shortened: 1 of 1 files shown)


docs/functions.md
```diff
@@ -0,0 +1,567 @@
...
+- 1: `[]any` - A slice of items to convert into an HCL list.
+
+Returns:
+
+- A string representation of the provided slice
+
+##### Example
+
+```
+[[ $cs := ( customSpew | withDisableCapacities true ) ]][[ $cs.Sdump . ]]
```
gulducat, line 514: should be a `toStringList` example

docs/functions.md
```diff
@@ -0,0 +1,567 @@
...
+##### Example
+
+```
+
+```
+
+##### Output
+
+```
+```
```
gulducat, line 98: empty example/output for `nomadRegions`

gulducat (approved):
couple lil things, but otherwise LGTM!

docs/functions.md
```diff
@@ -0,0 +1,567 @@
...
+
+```
+[[ $cs := ( customSpew | withSpewKeys true ) ]][[ $cs.Sdump . ]]
+```
+
+### Helper functions <a id="topicHelpers"></a>
+
+#### `fileContents` <a id="fileContents"></a>
+
+Imports the contents of a file on the local file system into the template at runtime.
```
josegonzalez, line 468: Can it be clarified that the files are relative to the directory where nomad-pack is invoked? I'm pretty sure they aren't relative to the job as that isn't working for me (which is frustrating, because thats where I'd expect them to be).

josegonzalez:
I think this is also missing the functions defined [here]https://github.com/hashicorp/nomad-pack/blob/3c0178a561b360906c591718edb25ae25ae9d964/internal/pkg/variable/parser/template_context.go#L159( for variable retrieval.

docs/functions.md
```diff
@@ -0,0 +1,567 @@
...
+- 2: `spew.ConfigState` - A customized Sprig printer created by `customSpew`
+
+##### Returns
+
+- The modified Sprig printer
+
+##### Example
+
+```
+[[ $cs := ( customSpew | withMaxDepth true ) ]][[ $cs.Sdump . ]]
```
SirHamburger, line 284: Max Depth should be an integer
```
[[ $cs := ( customSpew | withMaxDepth 5) ]][[ $cs.Sdump . ]]
```

docs/functions.md
```diff
@@ -0,0 +1,567 @@
...
+- 2: `spew.ConfigState` - A customized Sprig printer created by `customSpew`
+
+##### Returns
+
+- The modified Sprig printer
+
+##### Example
+
+```
+[[ $cs := ( customSpew | withDisableMethods true ) ]][[ $cs.Sdump . ]]
```
SirHamburger, line 306: withDisableMethods sets s.DisableMethods to true. So only "true" is not allowed (want 1 got 2). So it should be:
```
    [[ $cs := ( customSpew | withDisableMethods ) ]][[ $cs.Sdump . ]]
```

SirHamburger: Same with all the following, which use "true" as parameter

josegonzalez:
Could this get merged and then improved at a later date?

docs/functions.md
angrycub: The `fileContents` function is part of the job meta-templating and is run while **building** the job spec itself.  Should you want to import files while processing the jobspec , you should consider the HCL2 [`file`](https://developer.hashicorp.com/nomad/docs/job-specification/hcl2/functions/file/file) function, which would work more like you expect; but requires that the file to include always travel with the job spec.

PR #499, opened 2024-04-22
Detect os/arch more makily

gulducat:
My `linux/amd64` was being erroneously detected as `linux/arm64`, due to some subshell `|| ... &&` confusion.

This makes it more makey, which both works and reads a little cleaner IMO.


angrycub (approved):
:shipit:

PR #507, opened 2024-05-03
feat: output the modify index after planning

### LINKED ISSUE #506: nomad-pack plan does not output a check-index
Within CI pipelines it is common to split nomad plan and run phase with a manual gate in between to review and approve. 

In vanilla nomad the workflow goes something along the line of

`nomad plan job.hcl + make note of check-index -> review/approve -> nomad run -check-index $check-index job.hcl`

nomad-pack has implemented the check-index flag in its run command but does not emit during the plan phase leading to a somewhat clumsy workflow

`nomad-pack render job -> nomad plan job/job.hcl + make note of check-index -> review/approve -> nomad run -check-index $check-index job/job.hcl`

It would be awesome if nomad-pack could directly emit the check-index during its plan phase

attachmentgenie:
**Description**
This adds a stanza to the plan output that mentions the modify index for later use. Similar to how nomad implements it

fixes #506 

**Reminders**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 BUG FIXES:
 * runner: updated to hashicorp/nomad@1.7.2 to support `action` blocks [[GH-476](https://github.com/hashicorp/nomad-pack/pull/476)]
 
+IMPROVEMENTS:
+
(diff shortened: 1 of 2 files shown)


pkazmierczak (approved):
LGTM! Thanks, @attachmentgenie!

PR #566, opened 2024-07-31
build: move Darwin builds to Linux runner

tgross:
Pack doesn't require CGO and explicitly disables it in the build. We can build the Darwin builds on Linux and drop the expensive Darwin runners.

Ref: https://hashicorp.atlassian.net/browse/NET-9603


pkazmierczak (approved):
LGTM, thanks Tim!

PR #574, opened 2024-08-22
feat: add template function for reading files from the pack templates directory

josegonzalez:
**Description**

When distributing a pack, it is common to wish to ship larger config files as part of a pack in separate files. This allows for better readability, as well as ensuring users can fallback on syntax highlighting from their editors when working on those files.

**Reminders**


josegonzalez:
Is there anything I need to do to get this merged?

aleksanderllada:
This would be super useful for us too! Currently there's no way to inject configuration files without writing them directly inside a `template` block.

PR #581, opened 2024-10-16
build: update Nomad dependency to 1.9.0

jrasell:
Builds on the Go update branch, which is required by this change.

Nomad 1.9 removed support for hcl1. In order to update the Nomad
dependency, we too need to remove support for parsing templates
as HCL1.


pkazmierczak (approved):
LGTM, thanks for doing this @jrasell!

PR #651, opened 2025-03-17
sec: Update and suppress binary scanning failures.

jrasell:
`go-getter` is the source of the AWS library failure. This needs to be suppressed until that library is updated.

`golang.org/x/oauth2` has been updated to its latest version.

Related: https://github.com/hashicorp/go-getter/pull/467


pkazmierczak (approved):
LGTM!

PR #697, opened 2025-06-26
Required variables: error on unset variables that have no default

### LINKED ISSUE #695: Mechanism to mark Pack variables as required
In Terraform, a variable like this

```hcl
variable "message" {
  description = "The message to be logged"
  type        = string
  #default     = "" # no default
}
```

will cause an interactive prompt asking for the variable value, if it is not set with `-var` or `-var-file`.

I.e. it's assumed to be required if it has no default.

In Pack, the variable just ends up as `nil` without prompt or notification.  This results in strange contortions like this, down in template code, to explicitly fail if the var is not set.

```go
[[- define "message" -]]
[[- if not (var "message" .) -]]
[[- fail "must provide a message" -]]
[[- var "message" . -]]
[[- end -]]
[[- end -]]
```

and even then, the error message is rather verbose for someone who's just trying to use the pack:

```shell
$ nomad-pack render .
! Failed To Process Pack

    Error:   must provide a message
    Context:
        - Filename: testing/templates/_helpers.tpl
        - Position: 3,4
        - Registry Name: <<local folder>>
        - Pack Name: testing
        - Pack Ref: <<none>>
        - Pack Path: /home/dbennett/code/hashicorp/nomad-pack-community-registry/packs/testing
```

---

This should be a feature of the HCL/var parsing, rather than passing the buck down to template logic.

We could either
 1. Make these variables implicitly required -- error if they have no default and are not set
 2. Add a `required: bool` field to the `variable` block -- default `false` for backcompat

gulducat:
**Description**

With Nomad jobs, a `variable` without a `default` will produce an error during `nomad job run` if the variable is not set.

This PR aligns `nomad-pack` commands (render, plan, run) with that convention, removing the need for checking this in template logic.

<details><summary>Example results</summary>

Given this `variables.hcl`:

```hcl
variable "foo" {
  description = "i really need a foo"
  type        = string
}
variable "bar" {
  description = "i can use a bar"
  default     = "salad"
}
variable "bazzes" {
  description = "one or more baz"
  type        = list(string)
}
variable "good_luck" {
  default = {}
  type = object({
    some : number
  })
}
variable "who_knows" {}
```

```shell
$ nomad-pack info ./test/var-default
...
Pack "var-default" Variables:
        - "foo" (string: required) - i really need a foo
        - "bazzes" (list of string: required) - one or more baz
        - "who_knows" (unknown: required) -
        - "bar" (string: optional) - i can use a bar
        - "good_luck" (object: optional) 
```

The same errors from `render` below also happen for `plan` and `run`.

I could not find an easy way to make these errors less verbose, but I could hunt for it some more if desired.

```
$ nomad-pack render ./test/var-default
! Failed To Process Pack

    Error:   missing required variable: "who_knows"
    Context:
        - Registry Name: <<local folder>>
        - Pack Name: var-default
        - Pack Ref: <<none>>
        - Pack Path: /home/dbennett/code/hashicorp/nomad-pack/test/var-default

! Failed To Process Pack

    Error:   missing required variable: "foo" (i really need a foo)
    Context:
        - Registry Name: <<local folder>>
        - Pack Name: var-default
        - Pack Ref: <<none>>
        - Pack Path: /home/dbennett/code/hashicorp/nomad-pack/test/var-default

! Failed To Process Pack

    Error:   missing required variable: "bazzes" (one or more baz)
    Context:
        - Registry Name: <<local folder>>
        - Pack Name: var-default
        - Pack Ref: <<none>>
        - Pack Path: /home/dbennett/code/hashicorp/nomad-pack/test/var-default
```

```shell
$ nomad-pack generate var-file ./test/var-default
# variable "bar"
#   description: i can use a bar
#   default: "salad"
#
# bar="salad"


# variable "bazzes"
#   description: one or more baz
#   type: list(string)
#
# bazzes=«required»


# variable "foo"
#   description: i really need a foo
#   type: string
#
# foo=«required»


# variable "good_luck"
#   type: object({some = number})
#   default: {}
#
# good_luck={}


# variable "who_knows"
#
# who_knows=«required»



```

</details>

Since this is a breaking change, a CLI caller can fall back to the old logic with the new `--allow-unset-vars` flag or the `NOMAD_PACK_ALLOW_UNSET_VARS` environment variable.

Closes #695

**Reminders**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## UNRELEASED
 
+BREAKING CHANGES:
+* variables: Pack variables without `default` values now cause an error when not provided,
+  unless the `--allow-unset-vars` flag or `NOMAD_PACK_ALLOW_UNSET_VARS` env var are set.
+  [[GH-697](https://github.com/hashicorp/nomad-pack/pull/697)
+
 ## 0.3.0 (April 15, 2025)
 
 BREAKING CHANGES:
--- a/fixtures/v1/test_registry/packs/my_alias_test/deps/child2/variables.hcl
+++ b/fixtures/v1/test_registry/packs/my_alias_test/deps/child2/variables.hcl
@@ -9,6 +9,7 @@ variable "job_name" {
 
 variable "complex" {
   description = "complex object for rendering"
+  default     = {}
   type = object({
     name    = string
     address = string
--- a/fixtures/v2/test_registry/packs/my_alias_test/deps/child2/variables.hcl
+++ b/fixtures/v2/test_registry/packs/my_alias_test/deps/child2/variables.hcl
@@ -9,6 +9,7 @@ variable "job_name" {
 
 variable "complex" {
   description = "complex object for rendering"
+  default     = {}
   type = object({
     name    = string
     address = string
--- a/internal/cli/commands.go
+++ b/internal/cli/commands.go
@@ -63,6 +63,10 @@ type baseCommand struct {
 	// to variables defined in the pack should be ignored or produce an error
 	ignoreMissingVars bool
 
+	// allowUnsetVars suppresses errors from variables with nil values,
+	// i.e. those that are not set and have no default
+	allowUnsetVars bool
+
 	// autoApproved is true when the user supplies the --auto-approve or -y flag
 	autoApproved bool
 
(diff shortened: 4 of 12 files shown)


tgross (approved):
LGTM!

PR #737, opened 2025-09-11
Inject metadata into HCL templates

allisonlarson:
**Description**
Currently, nomad-pack injects metadata about the pack into the json definition of the job and not the HCL. When a pack job is managed outside of nomad-pack, like in the UI, the job loses its reference to the pack and is unable to be managed with nomad-pack until the job is completely removed. 

This change takes the rendered HCL and injects the metadata into the job/meta block, instead of just the json. It will merge with the defined meta attribute or block if it exists. In the case that it cannot successfully parse the HCL, the original job will be returned. If the job is has multiple meta attributes/blocks defined, it will merge with one and return the other untouched. This is invalid, but will be caught in the next step of the process, which is parsing the templates in nomad. The approach is to inject these values with as little change to the underlying job as possible, and allow any existing invalid HCL to still be passed through and caught in nomad during the parsing step. 

**Reminders**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 IMPROVEMENTS:
 * build: Update Go version to 1.24.6 [[GH-719](https://github.com/hashicorp/nomad-pack/pull/719)]
 * build: Update go-getter version to 1.7.9 [[GH-719](https://github.com/hashicorp/nomad-pack/pull/719)] to resolve CVE-2025-8959. Nomad Client Agents with Landlock support are not impacted by this vulnerability.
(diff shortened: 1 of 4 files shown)


tgross (dismissed):
Great work on this @allisonlarson! One minor question: what about the old code that was setting the `meta` attributes? That's redundant now so can we remove it?

allisonlarson:
@tgross Correct, it should be redundant! I had left it as-is in the case of a fallback, but it makes more sense to remove it completely.

tgross (approved):
LGTM! :shipit:

PR #749, opened 2025-10-07
Add preserve-resources flag to run command

allisonlarson:
**Description**
Adds the new nomad `-preserve-resources` flag to the `run` command, so that resources can be preserved on nomad-pack managed jobs. This is especially useful when using the nomad-autoscaler, so that whatever changes it made or recommends won't get written over when running a nomad-pack job. 

Internal ref: https://hashicorp.atlassian.net/browse/NMD-870

**Reminders**

none.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -1298,8 +1298,6 @@ github.com/hashicorp/net-rpc-msgpackrpc/v2 v2.0.1 h1:Y1sd8ZCCUUlUetCk+3MCpOwdWd+
 github.com/hashicorp/net-rpc-msgpackrpc/v2 v2.0.1/go.mod h1:wASEfI5dofjm9S9Jp3JM4pfoBZy8Z07JUE2wHNi0zuc=
 github.com/hashicorp/nomad v1.10.5 h1:KGQK0Ao9L7xtO/ZagS7tXD8MvFTr5R6y7SiOGhEXzWU=
 github.com/hashicorp/nomad v1.10.5/go.mod h1:NElc7qdOlCrkoKaJ8jyMo4+oX1U8TIxd8NM1TKtgVYE=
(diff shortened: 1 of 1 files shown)


go.sum
```diff
@@ -1300,6 +1300,8 @@ github.com/hashicorp/nomad v1.10.5 h1:KGQK0Ao9L7xtO/ZagS7tXD8MvFTr5R6y7SiOGhEXzW
 github.com/hashicorp/nomad v1.10.5/go.mod h1:NElc7qdOlCrkoKaJ8jyMo4+oX1U8TIxd8NM1TKtgVYE=
 github.com/hashicorp/nomad/api v0.0.0-20250630222842-3c2a6fefd3b2 h1:oEuBbu8IP4olIERuZalsO2pvrv7Tliilvjxy129oCJM=
 github.com/hashicorp/nomad/api v0.0.0-20250630222842-3c2a6fefd3b2/go.mod h1:y4olHzVXiQolzyk6QD/gqJxQTnnchlTf/QtczFFKwOI=
+github.com/hashicorp/nomad/api v0.0.0-20251006133510-26485c45a2fb h1:QdEQr0Vq3yU4dU/zN1lVZMHEgGsHPgFRP+tAGz0B6wU=
+github.com/hashicorp/nomad/api v0.0.0-20251006133510-26485c45a2fb/go.mod h1:sldFTIgs+FsUeKU3LwVjviAIuksxD8TzDOn02MYwslE=
```
gulducat, line 1304: seems odd that this is only additive. needa run `go mod tidy` which deletes the previous two lines.

gulducat (commented):
aside from needing a `go mod tidy`, lgtm!

just noting that this depends on hashicorp/nomad#26841, which made necessary server-side changes (in addition to the api package) and is not yet released.

what happens if someone tries to use this on an older server without those changes? is it just ignored?

allisonlarson:
@gulducat Thats correct, its ignored if its run against an older version of nomad. Is there a different way to handle dependencies like this? I was thinking that I could wait until its released and the version bumped, but I think the same scenario could still apply, right?

Issue #756, opened 2025-10-28
Namespace validation error fix and Improvement of deployment flow


Midas-sudo:
Hello everyone 👋 

So I've been trying to implement the use of nomad-pack on my company deployment flow and I've came across a bug as well as a possible feature. I discovered that the problem mentioned in this issue: https://github.com/hashicorp/nomad-pack/issues/522 which was causing nomad-pack to not detect some of my deployed jobs whenever they were deployed in a namespace that contained a dash.

My suggestion would be to improve the regex used to something like this:

```
namespaceRe := regexp.MustCompile(`(?m)namespace = \"([\w-]+)`)
regionRe := regexp.MustCompile(`(?m)region = \"([\w-]+)`)
```

This allows nomad-pack to correctly detect the namespace.

In adition i would suggest the implementation of a run option `--override-deploy` which tells nomad-pack to override the currently deploy job even if it was deployed through other methods.

Would like to hear the team feedback.
Is this something that could be implemented to nomad-pack? (I've currently developed this set of changes but dont seem to have contributions permission to open a PR in this repo)

jrasell:
Hi @Midas-sudo and thanks for raising this issue. I think both changes sound reasonable and I'll get this placed onto our backlog. If you did want to raise a PR from your work, you can do this via a fork of the repository.

Issue #522, opened 2024-06-14
Parsing of Region and Namespace fails when containing non-word characters.


johanforssell:
https://github.com/hashicorp/nomad-pack/blob/09ad3172cc07e8a545fa46ecb9c695b9341dc71c/internal/runner/job/job.go#L250-L251
`\w+` is used to parse the `region` and the `namespace`. This is inadequate when they contain non-word characters. 

"`-`" is a common non-word character which even shows up in the [official documentation of Namespaces](https://developer.hashicorp.com/nomad/tutorials/manage-clusters/namespaces#nomad-cli).
> $ export NOMAD_NAMESPACE=web-qa

On another note, the error message for failing to run/plan a job could use some improvements. I had to pull down the source code and do my own debugging to understand that my namespace was truncated.

slonopotamus:
This regex thing is very broken by design, see https://github.com/hashicorp/nomad-pack/issues/422#issuecomment-1715992565

PR #757, opened 2025-10-29
improves regex of namespace/regions and implements override-deploy flag

Midas-sudo:
**Description**
This PR improves the regex utilized for the namespace and region validation of the deployed/planned job. It allows the use of dashes in the name avoiding  the truncation of the name.

Also implements a flag to override current deployed jobs even if deployed through `nomad run` or other external deploy methods.

Fixes: #756
Fixes: #522

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## UNRELEASED
 
+IMPROVEMENTS:
+* cli: Add -deploy-override flag to plan and run commands to allow forcing deployment over existing jobs, including those not deployed by Nomad Pack [[GH-757](https://github.com/hashicorp/nomad-pack/pull/757)]
+* template: Improve regex parsing for namespace and region fields to support hyphenated values [[GH-757](https://github.com/hashicorp/nomad-pack/pull/757)]
+
 ## 0.4.1 (November 13, 2025)
 
 IMPROVEMENTS:


internal/runner/job/conflict.go
```diff
@@ -49,7 +49,7 @@ func (r *Runner) checkForConflict(jobName string) error {
 	// if there is a job with this name, that has no meta, it was
 	// created by something other than the package manager and this
 	// process should fail.
-	if existing.Meta == nil {
+	if existing.Meta == nil && !r.cfg.RunConfig.DeployOverride{
```
tgross: Rather than checking this flag multiple times, can't we check it once above all these checks and `return nil` if set?

tgross (changes requested):
Thanks for the contribution @Midas-sudo! I've edited your PR title and added `Fixes` trailers to point to the right issues.

I've left a couple notes. In addition:
* We'll need to add a changelog entry with a user-facing description of the changes
* In the future, if you're going to contribute a change it should be one PR per issue unless the issues are tightly related. It's fine for this PR.

Midas-sudo:
Hey @tgross :wave: 
Just did some adjustments to the PR according to your requests/sugestions.

Also is this fine for the changelog entries:

IMPROVEMENTS:
- cli: Add -deploy-override flag to plan and run commands to allow forcing deployment over existing jobs, including those not deployed by Nomad Pack [GH-757]
- template: Improve regex parsing for namespace and region fields to support hyphenated values [GH-757]

PR #787, opened 2025-12-15
CI: run tests on PRs

tgross:
The test workflow doesn't run on PRs, which is what allowed bugs like the one described in #786 to slip through to main.

Ref: https://github.com/hashicorp/nomad-pack/pull/786


gulducat (approved):
🧪

tgross:
Failing test is because of https://github.com/hashicorp/nomad-pack/pull/786

PR #791, opened 2026-01-16
variables: Add supported var file prefixes on error message.

### LINKED ISSUE #488: 1.0.1 gives `unrecognized file format suffix ".tfvars".` error
Running my packs that are working fine on 0.0.1-techpreview3, updating to 1.0.1 gives a new error
`Error: Cannot read from ../vars/do/prometheus.tfvars: unrecognized file format suffix ".tfvars".`

Not sure what else to provide to reproduce, file is a regular tfvars file with variables for the pack configured

jrasell:
This change adds a little more detail to the error returned to the user when a variable file is provided with an unsupported prefix. It should help identify the change needed.

Closes #488


pkazmierczak (approved):
LGTM!

Issue #278, opened 2022-07-26
change exit code from 127 to 0 for `nomad-pack`


ksatirli:
Hello maintainers 👋🏼 

When running `nomad-pack` and `nomad-pack help`, the exit code is `127`. 

For a successful operation like `help` (which `nomad-pack` with no commands defaults to), I'd expect this code to be `0`.

Is this something you'd consider changing?

Thanks!

PR #796, opened 2026-02-03
fix: return exit code 0 for help commands instead of 127

hashi-divyansh:
**Description**

**Problem:**
- Running 'nomad-pack' or 'nomad-pack help' returned exit code 127 instead of 0
- Exit code 127 typically means "command not found" in Unix/Linux
- Help commands should return 0 to indicate successful operation

**Fix:** 
- Convert mitchellh/cli exit code 127 to 0 in Main()
- Return exit code 1 for invalid arguments in help commands
- Refactor help commands to eliminate code duplication

[Jira Ticket](https://hashicorp.atlassian.net/browse/NMD-1184?atlOrigin=eyJpIjoiOTU0YTcwOTM5MTYwNDZlYmJjMzg2YjNjMGRiMmI2YmIiLCJwIjoiaiJ9)

[Fix Github Issue #278](https://github.com/hashicorp/nomad-pack/issues/278)

**Reminders**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ IMPROVEMENTS:
 * template: Improve regex parsing for namespace and region fields to support hyphenated values [[GH-757](https://github.com/hashicorp/nomad-pack/pull/757)]
 * variable: Improve the error message returned when using a variable file with an unsupported file extension [[GH-791](https://github.com/hashicorp/nomad-pack/pull/791)]
 
+BUG FIXES:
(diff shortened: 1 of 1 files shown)


internal/cli/main.go
```diff
@@ -100,6 +100,12 @@ func Main(args []string) int {
 		panic(err)
 	}
 
+	// The mitchellh/cli library returns 127 (cli.RunResultHelp) when displaying help.
+	// We want help to be a successful operation with exit code 0.
+	if exitCode == 127 {
+		return 0
+	}
```
pkazmierczak: I don't think we want that. If users submit a flag that doesn't exist, the CLI should return 127.

PR #797, opened 2026-02-04
fix: renamed cache pkg to caching

ritesh-harihar:
**Description**

- Renamed cache pkg to caching
- [JIRA Ticket](https://hashicorp.atlassian.net/browse/NMD-1153)

### DIFF
--- a/internal/cli/cli_test.go
+++ b/internal/cli/cli_test.go
@@ -53,7 +53,7 @@ const (
 )
 
 func TestCLI_CreateTestRegistry(t *testing.T) {
-	// This test is here to help setup the pack registry caching. It needs to be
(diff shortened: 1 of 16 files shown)


internal/cli/cli_test.go
```diff
@@ -53,7 +53,7 @@ const (
 )
 
 func TestCLI_CreateTestRegistry(t *testing.T) {
-	// This test is here to help setup the pack registry cache. It needs to be
+	// This test is here to help setup the pack registry caching. It needs to be
```
pkazmierczak: I admit that `caching` is perhaps a better package name, but registry cache is still a registry cache.

pkazmierczak (approved):
LGTM!

Issue #446, opened 2023-10-24
Add registry update command


mikenomitch:
Right now you have to re-run the add command to update a registry. Ideally you could just call update

PR #811, opened 2026-02-19
feat: Add registry update command

ritesh-harihar:
**Description**

Didn't touched the caching layer. The existing Cache.Add() in [add.go] already handles overwriting previously cached registries. The update command just calls [globalCache.Add()] with the same AddOpts — it's purely a CLI-level addition. So [registry update] is essentially a UX alias — same logic, different command name and messaging.

1) registry_update.go -> New file — RegistryUpdateCommand with Run, Flags, Synopsis, Help, autocomplete methods. Includes a check that the registry already exists before updating (returns error directing user to [registry add] if not).

2) main.go -> Added ["registry update"] to [commonCommands] list and registered RegistryUpdateCommand in the [commands] map.

3) help.go -> Added "registry update" entry to the helpText map.

4) registry_help.go -> Updated subcommand messages and synopsis to include "update". 

[Jira Ticket](https://hashicorp.atlassian.net/browse/NMD-1134)
Fixes: https://github.com/hashicorp/nomad-pack/issues/446

**Reminders**

### DIFF
--- a/internal/cli/registry_update_test.go
+++ b/internal/cli/registry_update_test.go
@@ -25,11 +25,13 @@ func TestCLI_RegistryUpdate_TooManyArgs(t *testing.T) {
 	must.StrContains(t, result.cmdOut.String(), "this command requires exactly 1 args")
 }
 
-// TestCLI_RegistryUpdate_NoArgs verifies that providing no arguments returns an error.
+// TestCLI_RegistryUpdate_NoArgs verifies that providing no arguments returns an error
+// and shows usage information.
 func TestCLI_RegistryUpdate_NoArgs(t *testing.T) {
 	result := runPackCmd(t, []string{"registry", "update"})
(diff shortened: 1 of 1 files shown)


internal/cli/registry_update.go
```diff
@@ -0,0 +1,196 @@
...
+		errorContext.Add(errors.UIContextPrefixRegistryTarget, c.target)
+	}
+
+	// Update the registry or registry target in the global cache
+	globalCache, err := caching.NewCache(&caching.CacheConfig{
+		Path:   caching.DefaultCachePath(),
+		Logger: c.ui,
+	})
+	if err != nil {
+		return 1
```
tgross: Does this throw out the error message and just exit 1? (Same with `Load`, `Add`, etc, below)

ritesh-harihar: Added useful error messages.

PR #812, opened 2026-02-19
feat(template): add nomadJobAllocations template function for output templates

### LINKED ISSUE #130: template: add Nomad job allocations template func
When creating output templates, it would be nice to be able to query the deployed job pack and detail information about the allocations. Additionally, if possible, the function could take an option list of status' which provides filtering of the job allocations.

vishnuedavarath:
**Description**

This PR adds a new template function `nomadJobAllocations` that enables pack authors to query and display allocation information in `outputs.tpl` after a pack is deployed.

Features:
- Query allocations by job ID
- Optional status filtering via list (e.g., "running", "pending")
- Retry mechanism (5 attempts, 500ms intervals) to handle timing
  window after job registration

**Usage Examples**
_Basic usage - all allocations:_

```hcl
[[ range nomadJobAllocations "my-job" ]]
  - ID: [[ .ID ]]
  - Status: [[ .ClientStatus ]]
  - Node: [[ .NodeName ]]
[[ end ]]
```
_With status filtering:_
```hcl
Running allocations:
[[ range nomadJobAllocations "my-job" (list "running") ]]
  - [[ .ID ]] on [[ .NodeName ]]
[[ end ]]
```
_Using pack variables:_
```hcl
[[ range nomadJobAllocations .my_pack.job_name ]]
  [[ .ID ]]: [[ .ClientStatus ]]
[[ end ]]
```

Fixes #130

**Reminders**

### DIFF
--- a/docs/functions.md
+++ b/docs/functions.md
@@ -98,6 +98,54 @@ default: Default shared namespace
 ```
 
 
+### Job functions
+
+#### `nomadJobAllocations` <a id="nomadJobAllocations"></a>
+
+The `nomadJobAllocations` function returns allocation information for a deployed job. This is useful in output templates (`outputs.tpl`) to display allocation details after a pack is deployed.
+
+##### Parameters
+
+- 1: `string` - The job ID to query allocations for
+- 2: (optional) `[]string` or `string` - Status filter(s) to limit results (e.g., "running", "pending", "complete", "failed")
+
+##### Returns
+
+- `error` or [`[]*api.AllocationListStub`][] containing allocation details for the job.
+
+##### Example
+
+Get all allocations for a job:
+
+```
+[[ range $alloc := nomadJobAllocations "my-job" -]]
+  - [[ $alloc.ID ]]: [[ $alloc.ClientStatus ]]
+[[ end -]]
+```
+
+Get only running allocations:
+
+```
+[[ range $alloc := nomadJobAllocations "my-job" "running" -]]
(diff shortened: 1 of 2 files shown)


tgross (commented):
This is a good first pass, but there's a lot of complexity to running a job that's not being covered here:
* If you're running a [periodic](https://developer.hashicorp.com/nomad/docs/job-specification/periodic) or [parameterized](https://developer.hashicorp.com/nomad/docs/job-specification/parameterized) batch job, it has no allocations right away.
* When a job is updated, the scheduler follows the [`update`](https://developer.hashicorp.com/nomad/docs/job-specification/update) block. So in cases where someone has `update.max_parallel=1`, this would immediately return as soon as the first alloc is created but not the rest of them. That's probably not what folks are looking for.
* When `service` or `system` jobs are run, they create a Deployment. You can monitor the deployment for health to get allocation info.

I think we want something more like how the `nomad job run` command works currently. When a job is registered, Nomad returns an evaluation ID. If you make a [blocking query](https://developer.hashicorp.com/nomad/api-docs#blocking-queries) on the evaluation, you'll be able to get back the list of allocations that it initially creates. And if there's a deployment to monitor, that ID will be in the eval as well.

vishnuedavarath:
Understood. I'll close this PR and take a different approach.

As you suggested, I'll implement an evaluation monitor similar to `nomad job run`, controlled by a `-detach` flag.

Thanks for the guidance, I'll open a fresh PR once I have something working.

PR #825, opened 2026-03-02
Stop nonexistent jobs in a deployed pack

ritesh-harihar:
**Description**
When a job is removed from a pack definition and the pack is re-deployed using nomad-pack run, the deleted job continues running in Nomad. This happens because the Deploy method only registers jobs currently defined in the pack templates and does not reconcile them against previously deployed jobs tied to the same pack.deployment_name.

In job.go file, the Deploy method iterates over r.parsedTemplates and registers each job with Nomad. However, it does not check for jobs that were previously deployed under the same pack.deployment_name but are no longer present in the current pack definition. As a result, removed jobs are left running.

Fix:-
Created reconcile.go with two methods: 

1) stopRemovedJobs — Called at the end of Deploy(). Builds a set of job IDs just deployed, queries Nomad for all jobs belonging to the same pack.deployment_name, and deregisters any that are no longer in the current pack definition. 

2) findStaleJobs — Queries the Nomad API for all running jobs, filters to those matching the deployment's pack.deployment_name metadata, and returns jobs whose IDs are not in the currently deployed set.

Modified job.go file to call r.stopRemovedJobs(ui, errorContext) after all current jobs have been successfully registered. 

The stale jobs are stopped (not purged) so they remain queryable via Nomad's garbage collection, matching the behavior of nomad-pack stop.

[Jira Ticket](https://hashicorp.atlassian.net/browse/NMD-1167)

**Reminders**

### DIFF
--- a/internal/runner/job/reconcile.go
+++ b/internal/runner/job/reconcile.go
@@ -42,86 +42,76 @@ func (r *Runner) stopRemovedJobs(ui terminal.UI, errorContext *errors.UIErrorCon
 	}
 
 	// Stop each stale job.
-	for _, staleJob := range staleJobs {
-		jobID := *staleJob.ID
+	for _, stub := range staleJobs {
 		ui.Warning(fmt.Sprintf(
 			"Job '%s' is no longer defined in pack deployment '%s' - stopping",
-			jobID, r.runnerCfg.DeploymentName,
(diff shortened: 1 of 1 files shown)


pkazmierczak (approved):
LGTM!

PR #841, opened 2026-03-16
fix: added validation block in variable schema

ritesh-harihar:
**Description**
This PR adds support for HCL2-style validation blocks inside variable declarations. Previously, variable decoding rejected nested validation blocks with: “Blocks of type validation are not expected here.”

Issue:
When HCL's body.Content(schema) is called in the decoder, it compares the actual file content against this schema. Since validation wasn't listed as an allowed block type, HCL raised the error.

What changed

1) Schema — Added validation as an allowed block type inside VariableBlockSchema, and defined a new ValidationBlockSchema requiring condition (expression) and error_message (string).

2) Variable struct — Added a Validations []Validation field and a Validation type holding the condition expression, error message, and source range.

3) Validation functions — Created validation_funcs.go exposing go-cty stdlib functions (contains, length, regex, regexall, comparison operators, etc.) so condition expressions can use them.

4) Decoder — After parsing description/type/default, the decoder now iterates any validation blocks in the content, parses each one's condition expression and error_message string, and appends them to v.Validations.

5) Variable.Validate() method — Evaluates each validation rule by building an hcl.EvalContext where var.<name> resolves to the variable's current value, then calls rule.Condition.Value(ctx). If the result is false, it emits a diagnostic with the custom error_message.

6) Both V1 and V2 parsers — After all overrides (env → file → CLI flags) are merged into root variables, both parsers now call v.Validate() on every variable and collect any failing diagnostics.

[JIRA Ticket](https://hashicorp.atlassian.net/browse/NMD-1185)

**Reminders**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## UNRELEASED
 
 IMPROVEMENTS:
+* cli: Add registry now honors default main/master branch [[GH-843](https://github.com/hashicorp/nomad-pack/pull/843)]
 * variable: Variables declaration now supports validation stanza [[GH-841](https://github.com/hashicorp/nomad-pack/pull/841)]
 
 ## 0.4.2 (March 16, 2026)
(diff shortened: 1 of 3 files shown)


pkazmierczak (approved):
LGTM!

Issue #473, opened 2023-11-17
[feature request] Make variable source pluggable


akamensky:
Currently variable need to be provided either on CLI or as a file. This may be a limiting factor to the uses of this tool. I think being able to have multiple sources for variables can be very beneficial to this tool.

As an example:
1. Pack that defines variables A, B and C
2. Consul KV has key value pairs for B and C
3. Running the pack with variables A and B and also pointing it to Consul KV as another variable source
4. The output will use A and B as provided from CLI and C as available in Consul KV (B on CLI considered an override)

As an alternative -- extend templating engine to be able to query some of other Hashicorp tools for the data, such as Consul KV, Nomad Variables and Vault secrets engines.

I think this request also supersedes #409 with a bit of higher-level/wider scope.

akamensky:
FWIW, for Consul KV I could now call `consul-template` to generate variable file from KV data and then use that file as an input for nomad-pack. But I think providing ability to directly externalize source of variables would play well with #472

PR #848, opened 2026-03-19
feat: Add vaultRead() and vaultKV() template functions to read secrets from HashiCorp Vault (GH-473)

DeekshithaTimmareddy:
**Description**

Implements Vault KV integration for nomad-pack templates, allowing packs to read secrets from HashiCorp Vault.

## Testing Steps

# Testing Environment Setup

1) Started Vault Dev Server
    
       vault server -dev

image

Here vault is running at : 'http://127.0.0.1:8200'
 and the root token is : 'hvs.W3kXXdCLhHibE3qMTQpYRhuS'

2) Configured Environment Variables
image

3) Verified Vault Status
image

# Testing Data Setup

Added test secrets to Vault KV v2:

image

verified secrets were stored:

image

here all secrets confirmed present in Vault

# Test Pack Creation:

1) Created a test pack from existing template:
 
image

2) Modified template to use Vault functions:

Added Vault function calls at the top:
 
[[ $apiKey := vaultRead "secret/data/api-key" ]]
[[ $dbData := vaultRead "secret/data/database" ]]

Added environment variables in task section:

env {
  API_KEY = [[ index $apiKey "api_key" ]]
  DB_URL = [[ index $dbData "url" ]]
  DB_PASSWORD = [[ index $dbData "password" ]]
}

# Test Execution

Built nomad-pack and rendered template with Vault integration:

image

Here output showed successful Vault secret retrieval

# Testing Backward Compatibility

image

Pack rendered successfully without Vault functions without errors

 
## Changes
- Added `vaultRead()` and `vaultKV()` template functions
- Automatic KV v1/v2 detection and data unwrapping
- Integration with CLI commands (run, plan, render)
- Configuration via standard Vault environment variables (`VAULT_ADDR`, `VAULT_TOKEN`)
- Complete documentation with usage examples
- Move vault/api to direct dependency

**Reminders**


jrasell (commented):
Hi @TIMMAREDDYDEEKSHITHA-ship-it and thanks for this. I like the idea of having the different functionality to get the unwrapped KV data or the raw object!

In #850 we are adding CLI flags as well as env var support to configure the Consul client and I think having that within this functionality will be useful too. It'll also be nice to have the UX match each integration as much as possible. The Vault config object has `ReadEnvironment` and `ConfigureTLS` functions which I think we should look at using to provide env var configuration and TLS support.

It would also be great to have some minimal tests to cover the Vault API functionality. I think there are a few possibilities here:
- have a basic test HTTP server that can be used to mock the Vault API and return data
- run Vault in dev mode via helper scripts, populate with data in the test
- check if Vault has a similar HTTP test agent that can be used in tests

docs/functions.md
```diff
@@ -130,6 +130,89 @@ The `nomadVariable` function retrieves a specific Nomad Variable by path and nam
...
+
+Currently only vault is supported as a variable source.
+
+Variable precedence is:
+
+1. external variable source
+2. environment variables
+3. variable files
+4. CLI --var
+This means CLI --var values override Vault-sourced values.
```
jrasell, line 214: I think this list order needs to be flipped as the numbered list implies external sources have the highest priority (listed as #1), whereas the closing sentence says "CLI `--var` values override Vault-sourced values", which means Vault is actually lowest priority.

internal/pkg/renderer/funcs.go
```diff
@@ -238,3 +244,40 @@ func withSpewKeys(s *spew.ConfigState) any {
...
+			return data, nil
+		}
+
+		//for KV v1 or other engines, return data directly
+		return secret.Data, nil
+	}
+}
+
+// vaultKV reads from Vault KV and returns raw data without unwrapping.
+// useful when you need access to metadata or want to handle KV versions yourself.
```
jrasell, line 271: The comment suggests we get the complete response including metadata, but the function is returning `secret.Data` which is the same as the `vaultRead` function without the v1/2 unwrapping. Do we need to update the comment or the function?

internal/cli/cli_test.go
```diff
@@ -88,6 +121,71 @@ func TestCLI_Version(t *testing.T) {
...
+		"--no-format=true",
+		"--var-source=vault",
+		testfixture.AbsPath(t, "v2/simple_raw_exec_v2"),
+	})
+
+	must.Eq(t, 1, result.exitCode)
+	out := result.cmdOut.String()
+	if !strings.Contains(out, "vault variable source requires a non-empty path") && !strings.Contains(out, "a Vault-backed variable source was requested, but no Vault client is configured") {
+		t.Fatalf("expected missing-path or missing-client error, got:\n%s", out)
+	}
```
jrasell, line 171: We could use ` must.StrContains` here and it's probably worth having an assertion for each string check, otherwise the test feels a little fragile and can pass if either error message appears.

internal/pkg/renderer/funcs.go
```diff
@@ -238,3 +244,40 @@ func withSpewKeys(s *spew.ConfigState) any {
 	s.SpewKeys = true
 	return s
 }
+
+// vaultRead reads a secret from Vault with automatic KV v1/v2 detection.
+// For KV v2, it automatically unwraps nested data structure.
+func vaultRead(client *vault.Client) func(string) (map[string]interface{}, error) {
+	return func(path string) (map[string]interface{}, error) {
+		secret, err := client.Logical().Read(path)
```
jrasell, line 252: The Vault API supports `ReadWithContext` in order to provide timeout or context cancellation. If Vault is slow or unreachable, template rendering and variable parsing will block indefinitely with no way to cancel.

internal/pkg/renderer/funcs.go
```diff
@@ -238,3 +244,40 @@ func withSpewKeys(s *spew.ConfigState) any {
...
+		//for KV v1 or other engines, return data directly
+		return secret.Data, nil
+	}
+}
+
+// vaultKV reads from Vault KV and returns raw data without unwrapping.
+// useful when you need access to metadata or want to handle KV versions yourself.
+func vaultKV(client *vault.Client) func(string) (map[string]interface{}, error) {
+	return func(path string) (map[string]interface{}, error) {
+		secret, err := client.Logical().Read(path)
```
jrasell, line 274: The Vault API supports `ReadWithContext` in order to provide timeout or context cancellation. If Vault is slow or unreachable, template rendering and variable parsing will block indefinitely with no way to cancel.

jrasell (changes requested):
Hi @TIMMAREDDYDEEKSHITHA-ship-it and thanks for the continued work on this.

One meta comment and item that might be worth discussing is whether it would ever be useful to have multiple variable sources? In the current implementation, it looks like it can only be vault or consul, not potentially Vault and Consul.

It would also be nice to have some tests covering the Vault API calls here as that is the bulk of the functionality.

DeekshithaTimmareddy:
Closing as it was already done !!

PR #850, opened 2026-03-23
feat: Add Consul KV integration for template variables (GH-473)

DeekshithaTimmareddy:
**Description**

Adds support for retrieving variables from Consul's Key-Value store in pack templates, addressing feature request #473.

This implementation adds two new template functions:
- `consulKey(key)` - Retrieves a single key-value pair
- `consulKeys(prefix)` - Retrieves multiple key-value pairs with a given prefix

Currently, variables must be provided via CLI or files, which can be limiting for dynamic configurations. This feature enables packs to fetch configuration directly from Consul KV, similar to the existing Nomad Variables integration.

## Changes

- Added `consulKey()` and `consulKeys()` template functions
- Added `ConsulClient` field to Renderer struct
- Updated CLI commands (`run`, `plan`, `render`) to support Consul KV flags:
  - `--consul-kv-address` - Consul server address
  - `--consul-kv-token` - Consul ACL token
  - `--consul-kv-namespace` - Consul namespace (Enterprise)
- Added comprehensive documentation in `docs/functions.md`
- Updated `go.mod` to include Consul API as direct dependency

## Testing Steps

1) Start Consul and checking if it is already running on Mac

image

2) Adding test data to Consul

image

3) Verifying data in Consul

image

4) Creating and verifying a test pack

image

Building the nomad-pack and verifying the flag exists

image

After creating variables.hcl and metadata.hcl files

image

Successful render

# Test - Error Handling

image

wrong port(9999) gave clear error which shows proper error handling when Consul is unavailable

**Reminders**


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 IMPROVEMENTS:
 * cli: Add registry now honors default main/master branch [[GH-843](https://github.com/hashicorp/nomad-pack/pull/843)]
 * variable: Variables declaration now supports validation stanza [[GH-841](https://github.com/hashicorp/nomad-pack/pull/841)]
+* template: Add `consulKey()` and `consulKeys()` functions to access Consul KV store for template variables [[GH-473](https://github.com/hashicorp/nomad-pack/pull/850)]
```
jrasell: We can link to the PR number and readers can navigate to the linked issue if desired.
```suggestion
* template: Add `consulKey()` and `consulKeys()` functions to access Consul KV store for template variables [[GH-850](https://github.com/hashicorp/nomad-pack/pull/850)]
```

internal/cli/plan.go
```diff
@@ -225,6 +298,73 @@ func (c *PlanCommand) Flags() *flag.Sets {
...
+                    Overrides CONSUL_HTTP_SSL_VERIFY environment variable.`,
+		})
+
+		f.StringVar(&flag.StringVar{
+			Name:    "consul-kv-tls-server-name",
+			Target:  &c.consulTLSServerName,
+			Default: "",
+			Usage: `Server name to use as SNI host when connecting via TLS. 
+                    Overrides CONSUL_TLS_SERVER_NAME environment variable.`,
+		})
```
jrasell: This initial feature targets Consul KV but the client could be used in the future for other Consul API calls. I think we should drop the `kv` specific notion from the CLI tags, so in the future it can be used for other API endpoints. The `nomad agent -h |grep consul` command could also be used to get an idea of the flags Nomad has to configure the Consul API client.

The flag setup is duplicated in multiple places, so we should use a helper function to generate the Consul flag set. This means if we need to change a change in the future, we can make it in a single place and reduces the amount of code we have to write and maintain.

internal/cli/plan.go
```diff
@@ -54,7 +68,66 @@ func (c *PlanCommand) Run(args []string) int {
...
+				consulConfig.TLSConfig.Address = c.consulTLSServerName
+			}
+		}
+
+		consulClient, err = consulapi.NewClient(consulConfig)
+		if err != nil {
+			c.ui.ErrorWithContext(err, "failed to create Consul client for KV operations", errorContext.GetAll()...)
+			return c.exitCodeError
+		}
+	}
```
jrasell: The Consul API default config uses environment variables to populate the configuration, so could use lean on the library to do this work for us, rather than have our own logic?

The client setup is duplicated in multiple places, so we should use a helper function to generate the Consul client. This means if we need to change a change in the future, we can make it in a single place and reduces the amount of code we have to write and maintain.

jrasell (changes requested):
Thanks for the changes @TIMMAREDDYDEEKSHITHA-ship-it! There is still a bit of code duplication I think we can tidy up. It would also be good to think about what tests we can add as part of this PR, as currently, we are adding ~600 LOC without one test.

CHANGELOG.md
TIMMAREDDYDEEKSHITHA-ship-it: Updated !!

internal/cli/plan.go
TIMMAREDDYDEEKSHITHA-ship-it: All changes have been tested locally with live consul instance, please let me know if there's anything else you'd like me to adjust !

internal/cli/helpers.go
```diff
@@ -559,3 +562,160 @@ func limit(s string, length int) string {
...
+		Default: "",
+		Usage: `Server name to use for TLS verification when communicating with Consul. 
+		        Can also be specified via the CONSUL_TLS_SERVER_NAME environment variable.`,
+	})
+}
+
+// LoadFromEnv loads configuration from environment variables.
+// Only loads values that haven't been set via command-line flags.
+// This implements the priority: CLI flags > environment variables.
+func (c *ConsulKVConfig) LoadFromEnv() {
```
jrasell: The `consulapi.DefaultConfig()` function already pull the standard `CONSUL_*` environment variables, so we should prefer relying on that rather than duplicating our own functionality here. Using that generated configuration as a base, we can then overlay CLI overrides when supplied.

internal/cli/helpers_consul_test.go
```diff
@@ -0,0 +1,138 @@
...
+// SPDX-License-Identifier: MPL-2.0
+
+package cli
+
+import (
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
```
jrasell: Let us the must library over testify.

jrasell (commented):
The Consul client setup is slightly duplicated in the `run`, `plan`, and `render` commands. It'll be good to use a single helper function that is called to generate a client if needed, that includes all the config parsing and precedence handling.

In the documentation it's probably a good idea to call out that the presence of the Consul address indicates Nomad Pack will attempt to generate an API client.

The manual testing steps are great, but this doesn't cover automated CI and adds additional manual overhead on all code review. It would also be great to have some minimal tests to cover the Consul API functionality. I think there are a few possibilities here:

- have a basic test HTTP server that can be used to mock the Consul API and return data
- run Consul in dev mode via helper scripts, populate with data in the test
- check if Consul has a similar HTTP test agent that can be used in tests

internal/cli/helpers.go
```diff
@@ -686,6 +689,186 @@ func limit(s string, length int) string {
...
+		"Skip TLS certificate verification when communicating with Consul. "+
+			"Can also be specified via the CONSUL_TLS_SKIP_VERIFY environment variable.")
+
+	flags.StringVar(&c.TLSServerName, "consul-tls-server-name", "",
+		"Server name to use for TLS verification when communicating with Consul. "+
+			"Can also be specified via the CONSUL_TLS_SERVER_NAME environment variable.")
+}
+
+// Package-level variable to track if Consul flags have been registered
+var consulFlagsRegistered = false
```
jrasell, line 742: This is a global variable that persists across all command instances within the same process. Because `run`, `plan`, and `render` all call `AddFlagsToSet`, only the first command to call it will actually register Consul flags. All subsequent calls silently return early. This also breaks any test that creates more than one command instance, since the flag state is never reset. We should remove the global and use per-`flag.Sets` or similar if needed.

internal/cli/helpers.go
```diff
@@ -686,6 +689,186 @@ func limit(s string, length int) string {
...
+	CACert        string // Path to CA certificate file
+	ClientCert    string // Path to client certificate file (for mTLS)
+	ClientKey     string // Path to client key file (for mTLS)
+	TLSSkipVerify bool   // Skip TLS certificate verification (for testing)
+	TLSServerName string // Override TLS server name
+}
+
+// AddFlags adds all Consul KV configuration flags to the provided flag set.
+// This method is called by run, plan, and render commands to register flags.
+func (c *ConsulKVConfig) AddFlags(flags *flag.FlagSet) {
```
jrasell, line 707: This function seems unused, so we can remove it and the import it creates.

internal/cli/helpers.go
```diff
@@ -686,6 +689,186 @@ func limit(s string, length int) string {
...
+	}
+
+	return consulapi.NewClient(cfg)
+}
+
+// getConsulClient creates a Consul API client if Consul is configured.
+// Returns nil client if no Consul address is configured (not an error).
+// The presence of a Consul address (via CLI flag or CONSUL_HTTP_ADDR env var)
+// indicates that Nomad Pack should attempt to create a Consul API client.
+func getConsulClient(consulKV *ConsulKVConfig, errorContext *errors.UIErrorContext, ui terminal.UI) (*consulapi.Client, error) {
```
jrasell, line 855: errorContext` and `ui` are accepted but never referenced in the function body. Should we be using them, or if not, lets remove them from the function signature.

internal/cli/helpers_consul_test.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/shoenig/test/must"
+)
+
+func TestConsulKVConfig_NewConsulClient(t *testing.T) {
+	tests := []struct {
+		name      string
+		envVars   map[string]string
+		config    ConsulKVConfig
+		wantAddr  string
+		wantToken string
```
jrasell, line 19: It doesn't look like these parameters are ever tested via an assertion. Is this something we want to add, or should we remove them?

internal/cli/helpers_consul_test.go
```diff
@@ -0,0 +1,135 @@
...
+			wantErr:   false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Set environment variables
+			for k, v := range tt.envVars {
+				os.Setenv(k, v)
+				defer os.Unsetenv(k)
```
jrasell, line 62: In tests we can use `t.Setenv(k, v)`: https://pkg.go.dev/testing#T.Setenv

jrasell (changes requested):
Thanks @TIMMAREDDYDEEKSHITHA-ship-it, I think this is getting close. Along with some inline comments, it would be really good to have some additional testing of the new code, including the template functions (via a mocked http server).

DeekshithaTimmareddy:
Closing as it was already done!

Issue #409, opened 2023-08-11
Add support for Nomad Variables


slonopotamus:
[Nomad Variables](https://developer.hashicorp.com/nomad/docs/concepts/variables)  is a lightweight alternative to Vault/Consul key-value storage with the added benefit of being scoped to Nomad namespaces.

jrasell:
Closing in favour of #473

slonopotamus:
No, this issue about a different thing. I want to be able to _create/update_ Nomad Variables via Nomad Pack. #473 wants to _read_ Nomad Variables when installing a pack.

PR #853, opened 2026-03-27
Add Nomad Variables support with automatic creation and deletion(Issue #409)

TIMMAREDDYDEEKSHITHA-ship-it:
**Description**

Implements Issue #409 - Adds support for Nomad Variables in nomad-pack, allowing packs to define persistent encrypted key-value pairs that are automatically created during pack deployment and deleted during pack destruction.

# Changes

Added nomad_variable block support in pack variables.hcl files
Automatic variable creation during nomad-pack run
Automatic variable deletion during nomad-pack stop
Full HCL parsing and validation for nomad_variable blocks

# Demonstration

Let us take an example file as variables.hcl

nomad_variable "app_config" {
  path = "nomad/jobs/myapp/config"
  namespace = "default"
  items = {
    database_url = "postgres://localhost:5432/mydb"
    api_key = "secret-key-123"
  }
}

nomad_variable "app_secrets" {
  path = "nomad/jobs/myapp/secrets"
  items = {
    admin_password = "super-secret"
  }
}

Then, 
1) Deploy Pack - Variables Created Automatically

image

2) Verifying variables exist

image

3) View Variable Contents

image

4) Variables Deleted on Pack Stop

image

Here, Automatic deletion via nomad-pack stop can work for registry packs but has path resolution issues with local packs 

**Reminders**

### DIFF
--- a/internal/cli/helpers.go
+++ b/internal/cli/helpers.go
@@ -18,6 +18,7 @@ import (
 	"github.com/hashicorp/nomad-pack/internal/pkg/variable/parser"
 	"github.com/hashicorp/nomad-pack/internal/runner"
 	"github.com/hashicorp/nomad-pack/internal/runner/job"
+	"github.com/hashicorp/nomad-pack/sdk/pack/variables"
 	"github.com/hashicorp/nomad-pack/terminal"
 )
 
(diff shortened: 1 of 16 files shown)


sdk/pack/variables/nomad_variables.go
```diff
@@ -0,0 +1,25 @@
...
+type NomadVariable struct {
+
+	// Name is the label from the nomad_variable block
+	Name string
+
+	// Path is where to store variable in Nomad
+	Path string
+
+	// Namespace is the Nomad namespace
+	Namespace string
```
hashi-divyansh, line 21: Maybe we should name it Namespaces []string so that users can add a variable to multiple namespaces at once 🤔. Please clarify it with someone senior.

hashi-divyansh (changes requested):

sdk/pack/variables/nomad_variables.go
jrasell: A pack can only be deployed at a single namespace at a time, so it feels like the Nomad variable should also only be written to that single namespace too. Let me know if you have thoughts on use-cases that warrant multiple namespace variable writes that I might be missing here.

internal/cli/helpers.go
```diff
@@ -199,6 +200,121 @@ func renderPack(
 	return r, nil
 }
 
+// createNomadVariables creates Nomad Variables defined in nomad_variable blocks
+func createNomadVariables(
+	parsedVars *parser.ParsedVariables,
+	client *api.Client,
+	ui terminal.UI,
+	errCtx *errors.UIErrorContext,
```
pkazmierczak: I don't think this is being used anywhere?

internal/cli/helpers.go
```diff
@@ -199,6 +200,121 @@ func renderPack(
...
+	ui.Success("Nomad Variables created successfully")
+	return nil
+}
+
+// deleteNomadVariables deletes Nomad Variables defined in nomad_variable blocks
+func deleteNomadVariables(
+	parsedVars *parser.ParsedVariables,
+	client *api.Client,
+	ui terminal.UI,
+	errCtx *errors.UIErrorContext,
```
pkazmierczak: this parameter doesn't seem to be used

internal/cli/helpers.go
```diff
@@ -199,6 +200,121 @@ func renderPack(
...
+		for _, nv := range nvList {
+			// Convert cty.Value items to map[string]string for Nomad API
+			items := make(map[string]string)
+			for key, val := range nv.Items {
+				// Convert cty.Value to Go interface
+				goVal, err := variables.ConvertCtyToInterface(val)
+				if err != nil {
+					return fmt.Errorf("failed to convert variable %s.%s: %w", nv.Name, key, err)
+				}
+				items[key] = fmt.Sprintf("%v", goVal)
```
pkazmierczak: I think this is potentially a problem or perhaps I don't fully understand how this works. 

`nomad_variable.items` can contain objects or maps, basically can be `any`. If we decode it and store it as a string, and then feed to Nomad API as a string, we might mess the formatting of the original string. I think it'd be best to serialize values as JSON?

internal/cli/run.go
```diff
@@ -143,6 +143,14 @@ func (c *RunCommand) run() int {
 		return 1
 	}
 
+	// create nomad variables
+	if r.ParsedVariables() != nil {
+		if err := createNomadVariables(r.ParsedVariables(), client, c.ui, errorContext); err != nil {
+			c.ui.ErrorWithContext(err, "failed to create Nomad Variables", errorContext.GetAll()...)
+			return 1
+		}
+	}
```
pkazmierczak, line 153: I think _maybe_ this is fine, but it's odd that this happens _after_ the templates are deployed. So it could be that we deploy a job successfully and then exit with non-zero code because we fail the variable creation. Not sure how to best handle this. I think it's ok because the alternative is a rollback, which I guess would have to be configurable, but we definitely have to call this out in the documentation.

pkazmierczak (commented):
Looks good! I left some comments, but I think we're getting there.

internal/cli/helpers.go
TIMMAREDDYDEEKSHITHA-ship-it: Fixed

internal/cli/helpers.go
TIMMAREDDYDEEKSHITHA-ship-it: Fixed!

internal/cli/helpers.go
TIMMAREDDYDEEKSHITHA-ship-it: yes! I used switch statement that properly serialises complex types as JSON using json.Marshal() while keeping simple types as-is.

internal/cli/run.go
TIMMAREDDYDEEKSHITHA-ship-it: Documented !!!

pkazmierczak (approved):
LGTM!

hashi-divyansh (approved):

Issue #852, opened 2026-03-27
Errors deploying jobs of type parameterized since 0.4.2


agarfer1:
### Environment

The error seems to be present since version 0.4.2 and is currently present in the nightly version.

The environment is a Gitlab pipeline using the dockerhub version of nomad-pack (**hashicorp/nomad-pack:0.4.2**)

### Description of the problem

This error occurs for jobs of type parameterized/dispatch. And it seems that since 0.4.2 nomad-pack tries to stop the jobs currently running. And it actually goes through with it and stops them.

```
Job 'job-requester' in pack deployment 'skeleton' registered successfully
warning: Job 'job-requester/periodic-1774607280' is no longer defined in pack deployment 'skeleton' - stopping
```

And at the end, this is reported as a failed deployment.

The same job definition but using version 0.4.1 works correctly:

```
Job 'job-requester' in pack deployment 'skeleton' registered successfully
Pack successfully deployed. Use packs/skeleton/ to manage this deployed instance with plan, stop, destroy, or info
Congrats! You deployed the xyz pack on Nomad.
```

### Reproduction steps

1. Create a job of type parameterized. Ideally the job should be some kind of infinite loop, for testing purposes.
2. Deploy the job. It might work if it has no instances of the job running.
3. Launch an instance of the job.
4. Perform the deploy again using **nomad-pack 0.4.2**. The process should fail.
5. Launch another instance of the job, since it would have been stopped.
6. Perform the deploy again using **nomad-pack 0.4.1**. It should work.

PR #855, opened 2026-03-30
Fix: Errors deploying jobs of type parameterized since 0.4.2

ritesh-harihar:
**Description**
This PR fixes a regression where nomad-pack run could stop active child jobs (for example, dispatched or periodic runs) during reconciliation. The issue appeared after stale-job cleanup was introduced and affected deployments where child jobs inherit pack metadata.

Root Cause
Reconciliation in reconcile.go lists jobs by pack.deployment_name and treats jobs not in the current rendered root job set as stale. Child jobs created from a root pack job were included in that list and misclassified as stale, which caused unintended Deregister calls. The previous logic did not robustly distinguish root jobs from generated child runs.

Fix
- Tightened stale-job selection to skip child jobs before stop actions.
- Child detection now uses ordered signals:
       - ParentID (preferred),
       - pack.job metadata consistency (pack.job != job ID means child),
       - slash-shape fallback only when pack.job metadata is missing.(Currently slash-based child suffixes are  hardcoded and used in nomad)                                         
- Added nil-safety in the reconcile loop to avoid dereference risk on unexpected list entries.

Why this approach
- Avoids hardcoding only periodic-* / dispatch-* assumptions.
- Keeps behavior future-tolerant without adding extra Nomad API lookups.
- Preserves original stale-job cleanup intent for true removed root jobs.

[ISSUE](https://github.com/hashicorp/nomad-pack/issues/852)

Testing:
- Before fix:
Screenshot 2026-03-30 at 4 03 07 PM

- After fix:

Screenshot 2026-03-30 at 7 18 38 PM

Screenshot 2026-03-30 at 5 09 19 PM

**Reminders**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 IMPROVEMENTS:
 * cli: Add registry now honors default main/master branch [[GH-843](https://github.com/hashicorp/nomad-pack/pull/843)]
-
+* cli: Fixed stale-job reconciliation so nomad-pack run no longer stops active parameterized/periodic child jobs [[GH-855](https://github.com/hashicorp/nomad-pack/pull/855)] 
 * variable: Variables declaration now supports validation stanza [[GH-841](https://github.com/hashicorp/nomad-pack/pull/841)]
 
 ## 0.4.2 (March 16, 2026)


internal/runner/job/reconcile.go
```diff
@@ -117,3 +127,36 @@ func (r *Runner) findStaleJobs(currentJobIDs map[string]struct{}) ([]*api.JobLis
...
+		return packJob != stub.ID
+	}
+
+	// Fallback only when pack.job metadata is missing.
+	return isChildStyleJobID(stub.ID)
+}
+
+// isChildStyleJobID is a conservative fallback: child jobs are represented as
+// "<parent>/<child-run-id>", so they always contain a non-leading slash.
+func isChildStyleJobID(jobID string) bool {
```
jrasell: It is possible Nomad job IDs are configured with a number of slashes irregardless of whether they are dispatched. I am curious of your thoughts on removing this fallback altogether and relying on the `ParentID` and metadata identification? Are there cases where this wouldn't be enough?

```console
$ nomad status
ID                      Type     Priority  Status   Submit Date
example/self/namespace  service  50        pending  2026-03-31T10:07:51+01:00
```

internal/runner/job/reconcile.go
```diff
@@ -117,3 +127,36 @@ func (r *Runner) findStaleJobs(currentJobIDs map[string]struct{}) ([]*api.JobLis
...
+	// Preferred signal from Nomad.
+	if stub.ParentID != "" {
+		return true
+	}
+
+	// Metadata-based signal from nomad-pack tags. For child jobs, pack.job points
+	// to the parent/root job name, which differs from the child ID.
+	if packJob, ok := stub.Meta[PackJobKey]; ok && packJob != "" {
+		return packJob != stub.ID
+	}
```
jrasell: Is `stub.Meta` always non-nil? If we are not sure, it might be worth adding a nil check.

ritesh-harihar: Added nil check.

internal/runner/job/reconcile.go
ritesh-harihar: “ParentID + pack.job " covers normal Nomad/nomad-pack behavior and removes slash-based false positives. 

The remaining edge case is when both signals are absent/inconsistent in stubs; **if we observe that in practice**, we can add a targeted Jobs().Info check only for ambiguous entries to keep reconciliation safe.

So instead of calling Jobs().Info for every job, call it only for uncertain ones.

Pseudo flow:

if stub.ParentID != "" → child
else if pack.job exists:
        pack.job != stub.ID → child
        pack.job == stub.ID → root
else (ambiguous) → call Jobs().Info:
        full ParentID != "" → child
        else → root


But not sure if we really need this API call?

ritesh-harihar:
> LGTM; I think we just need a changelog entry before merging.

Updated the changelog.

PR #875, opened 2026-04-23
fix: Output template rendering fails with --parser-v1 flag

### LINKED ISSUE #525: failed to render output template using nomad-pack run command
Hi there!

We updated to **nomad-pack** version **0.1.2** and now we're getting the error:

```
  Job 'example-job' in pack deployment 'example' registered successfully
Pack successfully deployed. Use ./packs/example to manage this deployed instance with plan, stop, destroy, or info
! Failed To Render Output Template
!   Error: failed to render outputs.tpl: template: outputs.tpl:1:6: executing "outputs.tpl" at <.my.job_name>: nil pointer evaluating parser.PackContextable.job_name
!   Context:
!     Pack Name: example
Cleaning up project directory and file based variables

ERROR: Job failed: exit code 1
```

Meanwhile the deployment is going well.

The problem does not occur when calling `nomad-pack render --parser-v1`, but does occur when calling `nomad-pack run --parser-v1` 
This has been present in **nomad-pack** since version **0.1.0**, in which the `--parser-v1` flag appeared.

Сontents of the `outputs.tpl` file:
```
[[ .my.job_name ]]:[[ .my.version ]] has been successfully deployed.

There are [[ .my.count ]] instances of your job now running on Nomad in [[ .my.namespace ]] namespace.
```

DeekshithaTimmareddy:
**Description**
Fixes #525

## Problem
Output template rendering fails when using --parser-v1 flag with error: "nil pointer evaluating parser.PackContextable.job_name"

image

## Solution
Added parser version check to use correct template context:

V1 parser: ConvertVariablesToMapInterface()
V2 parser: ToPackTemplateContext()

## Changes
Modified: internal/pkg/renderer/renderer.go (added version check)
Added: internal/pkg/renderer/renderer_output_test.go (5 tests)

### Template Rendering with V1 Parser
image

### Job Deployment success
image

image

**Reminders**

### DIFF
--- a/internal/pkg/renderer/renderer.go
+++ b/internal/pkg/renderer/renderer.go
@@ -189,9 +189,28 @@ func (r *Renderer) RenderOutput() (string, error) {
 		return "", err
 	}
 
-	ptc, _ := r.pv.ToPackTemplateContext(r.pack)
+	// Choose the appropriate template context based on parser version
+	// V1 parser uses flat map structure(e.g., .packname.variable)
+	// V2 parser uses nested structure with pack metadata (e.g., .my.variable)
+	var templateData interface{}
+	var diags hcl.Diagnostics
+
+	if r.pv.IsV1() {
+		// Use V1 conversion for --parser-v1 flag
+		templateData, diags = r.pv.ConvertVariablesToMapInterface()
+	} else {
+		// Use V2 conversion for default parser (current behavior)
+		templateData, diags = r.pv.ToPackTemplateContext(r.pack)
+	}
+
+	// Check for conversion errors
+	if diags.HasErrors() {
+		return "", fmt.Errorf("failed to create template context: %v", diags)
+	}
+
+	// Execute the template with the correct context
 	var buf strings.Builder
-	if err := r.tpl.ExecuteTemplate(&buf, r.pack.OutputTemplateFile.Name, ptc); err != nil {
+	if err := r.tpl.ExecuteTemplate(&buf, r.pack.OutputTemplateFile.Name, templateData); err != nil {
 		return "", fmt.Errorf("failed to render %s: %w", r.pack.OutputTemplateFile.Name, err)
 	}
 
--- a/internal/pkg/renderer/renderer_output_test.go
+++ b/internal/pkg/renderer/renderer_output_test.go
@@ -1,0 +1,202 @@
+// Copyright IBM Corp. 2023, 2026
+// SPDX-License-Identifier: MPL-2.0
+
+package renderer
+
+import (
+	"testing"
+	"text/template"
+
+	"github.com/hashicorp/nomad-pack/internal/pkg/variable/parser"
+	"github.com/hashicorp/nomad-pack/sdk/pack"
+	"github.com/hashicorp/nomad-pack/sdk/pack/variables"
+	"github.com/stretchr/testify/require"
+	"github.com/zclconf/go-cty/cty"
+)
+
+// TestRenderOutput_V1Parser_Success tests that output template rendering
+// works correctly with V1 parser. This is the fix for issue #525.
+func TestRenderOutput_V1Parser_Success(t *testing.T) {
+	// Create a pack with output template using V1 syntax
+	p := &pack.Pack{
+		Metadata: &pack.Metadata{
+			Pack: &pack.MetadataPack{
+				Name: "testpack",
+			},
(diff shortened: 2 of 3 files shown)


hashi-divyansh (commented):
.

internal/pkg/renderer/renderer_output_test.go
```diff
@@ -0,0 +1,202 @@
...
+package renderer
+
+import (
+	"testing"
+	"text/template"
+
+	"github.com/hashicorp/nomad-pack/internal/pkg/variable/parser"
+	"github.com/hashicorp/nomad-pack/sdk/pack"
+	"github.com/hashicorp/nomad-pack/sdk/pack/variables"
+	"github.com/stretchr/testify/require"
```
hashi-divyansh: For the new test cases use `github.com/shoenig/must `

internal/pkg/renderer/renderer_output_test.go
```diff
@@ -0,0 +1,202 @@
...
+	"github.com/zclconf/go-cty/cty"
+)
+
+// TestRenderOutput_V1Parser_Success tests that output template rendering
+// works correctly with V1 parser. This is the fix for issue #525.
```
hashi-divyansh:
```suggestion
// TestRenderOutput_V1Parser_Success tests that output template rendering
// works correctly with V1 parser.
```

internal/pkg/renderer/renderer_output_test.go
```diff
@@ -0,0 +1,202 @@
...
+	p := &pack.Pack{
+		Metadata: &pack.Metadata{
+			Pack: &pack.MetadataPack{
+				Name: "testpack",
+			},
+			App: &pack.MetadataApp{},
+		},
+		OutputTemplateFile: &pack.File{
+			Name:    "outputs.tpl",
+			Content: []byte("Job: {{ .testpack.job_name }}"),
```
hashi-divyansh: nomad-pack templates use `[[  ]] `syntax

DeekshithaTimmareddy: Done. Used demiliters !

internal/pkg/renderer/renderer.go
```diff
@@ -189,9 +189,28 @@ func (r *Renderer) RenderOutput() (string, error) {
...
+	var templateData interface{}
+	var diags hcl.Diagnostics
+
+	if r.pv.IsV1() {
+		// Use V1 conversion for --parser-v1 flag
+		templateData, diags = r.pv.ConvertVariablesToMapInterface()
+	} else {
+		// Use V2 conversion for default parser (current behavior)
+		templateData, diags = r.pv.ToPackTemplateContext(r.pack)
+	}
```
hashi-divyansh, line 204: Ok. This is the main fix 👍.

internal/pkg/renderer/renderer_output_test.go
hashi-divyansh, line 1: I ran the new test file against `main branch`  (without the fix) to verify which tests would actually catch the regression. Only 1 `TestRenderOutput_V1Parser_Success` failed  and all 5 were passed. Then I did deeper and relies the tests are missing the real production behaviour. 

In production, `Render()` is always called before `RenderOutput()` The tests manually construct `Renderer{} ` and bypass `Render()` entirely.  The test that calls Render() first and then `RenderOutput()` would replicate the actual call chan and be much stronger regression guard.

hashi-divyansh (commented):
Fix looks good 👍. I left a comment about strengthening the new tests. 

This PR contains multiple library version updates that don't seem related to the fix. It might be better to handle them in a separate PR, unless you think Dependabot won't catch them.

DeekshithaTimmareddy:
> Fix looks good 👍. I left a comment about strengthening the new tests.
> 
> This PR contains multiple library version updates that don't seem related to the fix. It might be better to handle them in a separate PR, unless you think Dependabot won't catch them.

Thanks for the review!

Removed Dependabot commits.

internal/pkg/renderer/renderer_output_test.go
DeekshithaTimmareddy: Thanks for the detailed feedback! I've updated tests

All tests now follow production flow. 4 V1 parser tests catch regression and 2 V2 parser tests validate V2 stilll works.

hashi-divyansh (approved):
LGTM

tgross (approved):
LGTM

PR #891, opened 2026-05-20
fix: Support branch names with slashes in --ref flag and add documentation

### LINKED ISSUE #321: Should be support branch for normal maintenance flow
Currently, nomad-pack supports the following options:

      --ref=<string>
        Specific git ref of the registry or pack to be added. Supports tags,
        SHA, and the latest. If no ref is specified, it defaults to the latest. Running
        "nomad registry add" multiple times for the same ref is idempotent,
        however running "nomad-pack registry add" without specifying a
        ref, or when specifying @latest, is destructive, and will overwrite
        current @latest in the global cache. Using ref with a file path is not
        supported.

      --target=<string>
        A specific pack within the registry to be added.
        
But if your updating a template, you do not want to directly work on main but often in a branch. So adding the support for --branch or allow --ref to also try to go to the branches would be nice.

Example:
nomad-pack registry add community github.com/hashicorp/nomad-pack-community-registry  --branch=feature/DEV-26331-add-templates

DeekshithaTimmareddy:
**Description**

Fixes #321 

This PR documents that '--ref' flag already supports branch names and eliminating the need for separate '--branch' flag. So documented in detail

# Testing

1) Main Branch

image

2) Non-default Branch (if exists)

image

3) Commit SHA

image

image

# Added support for branch names containing slashes

Before fix:

image

After fix:

image

# Fixed logging Display Bug

Before fix:

image

After fix:

image

**Reminders**


hashi-divyansh (changes requested):
Before going further with this, worth checking the @jrasell  [comment ](https://github.com/hashicorp/nomad-pack/issues/321#issuecomment-3759592860)on the issue.

This means the `--branch` flag itself **isn't needed** — `--ref` already handles branch names under the hood.

Please validate/test it from your end, and If it is true then we have to only update the docs and test cases.

jrasell:
> Before going further with this, worth checking the @jrasell [comment ](https://github.com/hashicorp/nomad-pack/issues/321#issuecomment-3759592860)on the issue.
> 
> This means the `--branch` flag itself **isn't needed** — `--ref` already handles branch names under the hood.
> 
> Please validate/test it from your end, and If it is true then we have to only update the docs and test cases.

Looking at the original request issue and my comment, it looks like the `-ref` flag can be used to either supply a Git SHA or a branch. In that case, users can either point to a specific commit in the repo, or a branch name to deploy which I believe covers most/all use cases. I think for the linked issue, we should expand the documentation to detail the ways in which the `-ref` flag can be used and not currently add a new CLI flag.

DeekshithaTimmareddy:
Thanks for the feedback @hashi-divyansh and @jrasell . 

I carefully tested it and found that --ref flag is already existing. So documented it instead of adding new --branch flag and also the existing tests in internal/pkg/caching/cache_test.go already test the --ref mechanism, so don't need to add any new test files.

docs/detailed-usage.md
```diff
@@ -67,6 +67,72 @@ To download single pack or an entire registry at a specific version/SHA, use the
...
+```
+nomad-pack registry add myregistry github.com/org/repo --ref=abc123def456
+```
+
+#### Known Limitations
+
+**Branch Names with Slashes:**
+
+Branch names containing slashes (e.g., `feature/add-templates`, `bugfix/issue-123`) are not supported due to go-getter URL parsing limitations. The tool will fail with an error like: `invalid ref: "feature"`.
+
```
hashi-divyansh: This limitation looks suspicious to me. Having `/ ` in branch names are pretty common and `go-getter` is quite matured library.  Please re-check it.

DeekshithaTimmareddy: Yes, this is not a limitation. Here the issue is branch names with slashes are not being URL-encoded before passing to go-getter. So I fixed it by adding 'url.QueryEscape()' and also I fixed a logging display bug and added 5 new tests for refs with slashes.

hashi-divyansh: So now we should remove these doc comment?

DeekshithaTimmareddy: Yes, I did remove

hashi-divyansh (dismissed):
Please also update the documentation for  `registry_update.go `

hashi-divyansh (dismissed):
LGTM!

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ IMPROVEMENTS:
 * variable: Variables declaration now supports validation stanza [[GH-841](https://github.com/hashicorp/nomad-pack/pull/841)]
 * variable: Add support for `nomad_variable` blocks with automatic lifecycle management during pack deployment and destruction [[GH-409](https://github.com/hashicorp/nomad-pack/pull/853)]
 * variable: Fixed variable file overrides (last supplied wins) [[GH-851](https://github.com/hashicorp/nomad-pack/pull/851)]
+* cli: Fixed `--ref` flag to support branch names with slashes and added documentation [[GH-891](https://github.com/hashicorp/nomad-pack/pull/891)]
```
jrasell: Does this need to go into the bug fixes section, rather than the improvements given the description?

internal/pkg/caching/add.go
```diff
@@ -145,7 +146,10 @@ func (c *Cache) cloneRemoteGitRegistry(opts *AddOpts) (string, error) {
 	logger := c.cfg.Logger
 	url := buildGoGetterGitURL(opts)
 
-	logger.Debug(fmt.Sprintf("go-getter URL is %s", url))
+	// Escape % signs in URL to prevent fmt.Sprintf interpretation in terminal layer
+	// %2F (URL-encoded slash) would be interpreted as format verb %2F otherwise
+	escapedURL := strings.ReplaceAll(url, "%", "%%")
+	logger.Debug("go-getter URL is " + escapedURL)
```
jrasell, line 152: I think this will produce `%%2F` escaping and we can keep the original as it shouldn't be interpreting `%2F` inside the argument as a format verb.

internal/pkg/caching/add.go
```diff
@@ -169,18 +173,21 @@ func (c *Cache) cloneRemoteGitRegistry(opts *AddOpts) (string, error) {
...
-		url = fmt.Sprintf("%s.git//packs/%s", src, opts.PackName)
+		urlStr = fmt.Sprintf("%s.git//packs/%s", src, opts.PackName)
 	}
 
 	if !opts.IsLatest() {
-		url = fmt.Sprintf("%s?ref=%s", url, opts.Ref)
+		// URL-encode the ref to handle branch names with special characters like slashes
+		// e.g., "feature/add-templates" becomes "feature%2Fadd-templates"
+		encodedRef := url.QueryEscape(opts.Ref)
+		urlStr = fmt.Sprintf("%s?ref=%s", urlStr, encodedRef)
```
jrasell, line 187: I think go-getter should handle slashes correctly without this and the [ref is be pulled from the request](https://github.com/hashicorp/go-getter/blob/6844d1e6916ad0e02c2c3feb65d8e2cbc94e0f68/get_git.go#L71) which results in the unescaped value anyway before being passed onto the git call.

jrasell (commented):
This looks great @TIMMAREDDYDEEKSHITHA-ship-it and the docs are really nice and clear. I've added a couple of inline notes, as I think go-getter can be left to do the lifting for us here.

internal/pkg/caching/add.go
DeekshithaTimmareddy: I tested both the scenarios.

Without %% escaping:
image

With %% escaping:
image

Here, logger internally uses fmt.Sprintf, so it converts %% to % in the output. I think this escaping is necessary. What do you say?

internal/pkg/caching/add.go
DeekshithaTimmareddy: You're correct that "q.Get("ref")" URL decodes the parameter, but the problem occurs during URL parsing before that.

Tested without encoding:
image

Here, '/' gets misinterpreted as a path separator during URL parsing, before 'q.Get("ref")' runs. So, I think this fix is necessary.

After fix:
image

CHANGELOG.md
DeekshithaTimmareddy: yes, changed!

jrasell (approved):

PR #892, opened 2026-05-21
Preserve pack metadata for jobs with top-level blocks like locals

gautamg795:
**Description**
Nomad Pack injects `pack.*` metadata into rendered job files so later `plan`, `run`, `stop`, and `destroy` commands can identify jobs managed by the same pack deployment.

The metadata injection path currently parses the rendered file with a schema that only permits top-level `job` blocks. If a valid Nomad job file also contains another top-level block, such as `locals`, that parse reports diagnostics and Nomad Pack skips metadata injection. The job can still be submitted, but it loses the `pack.*` metadata and future Nomad Pack operations treat it as unmanaged.

This changes the metadata injection parser to use `PartialContent`, so it can find top-level `job` blocks while preserving unrelated top-level HCL blocks. A regression test covers a job file with top-level `locals`.

**Reminders**

> Are there any changes to security controls (access controls, encryption, logging) in this pull request? If so, explain.

No. This does not change access controls, encryption, logging, or other security controls.


DeekshithaTimmareddy (approved):
nice fix, LGTM!

jrasell (approved):

PR #894, opened 2026-05-29
feat: add variable source interface and registry for pluggable architecture (Issue #473 - Subtask 1)

DeekshithaTimmareddy:
**Description**

Implements Subtask 1 of [Issue #473 ]

Creates the foundational architecture for pluggable variable sources, enabling nomad-pack to support multiple variable sources (environment, files, CLI, and future external sources like Consul KV, Vault, and Nomad Variables) with priority-based resolution.

# What Changed
New Components

1) VariableSource Interface (source.go)
Defines contract for all variable sources
Methods: Name(), Priority(), Fetch()
Enables pluggable architecture

2) Registry Pattern (registry.go)
Manages multiple sources with priority-based resolution
Validates sources (nil check, duplicate names)
Context-aware with cancellation support

3) BaseSource Implementation (base_source.go)
Eliminates code duplication for env/file/CLI sources
Handles edge cases (nil vars, empty maps, context cancellation)
Provides convenience constructors

4) Parser Integration (parser_v2.go)
Replaces manual merge loops with registry
Adds 30-second timeout protection

**Reminders**


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ IMPROVEMENTS:
 * variable: Variables declaration now supports validation stanza [[GH-841](https://github.com/hashicorp/nomad-pack/pull/841)]
 * variable: Add support for `nomad_variable` blocks with automatic lifecycle management during pack deployment and destruction [[GH-409](https://github.com/hashicorp/nomad-pack/pull/853)]
 * variable: Fixed variable file overrides (last supplied wins) [[GH-851](https://github.com/hashicorp/nomad-pack/pull/851)]
+* variable: Add pluggable variable source architecture for multi-source support [[GH-894](https://github.com/hashicorp/nomad-pack/pull/894)]
```
tgross: This isn't a user-facing change (yet), so it doesn't belong in the changelog.

internal/pkg/variable/source/source_test.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	source := &mockSource{
+		name:     "mock",
+		priority: 10,
+		vars:     []*variables.Variable{testVar},
+	}
+
+	require.Equal(t, "mock", source.Name())
+	require.Equal(t, 10, source.Priority())
```
tgross: This repo has `shoenig/test`; you should be writing new tests using that. 

(Or decide as a team you're going to rip it out; `shoenig/test` is strictly better IMO and fixes serious correctness bugs in `testify`, but go with one or the other instead of mixing.)

internal/pkg/variable/source/source_test.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func (m *mockSource) Fetch(ctx context.Context, packID pack.ID) ([]*variables.Variable, error) {
+	if m.err != nil {
+		return nil, m.err
+	}
+	return m.vars, nil
+}
+
+func TestVariableSource_Interface(t *testing.T) {
```
tgross: Every test you write should be able to justify the time it takes to run on every PR for the rest of the lifetime of the project (or until the feature is removed), and the developer time it takes to keep it up to date over time. This test _only_ exercises the behavior of the mock, and Go language constructs around interfaces. Don't write tests like this!

internal/pkg/variable/source/source.go
```diff
@@ -0,0 +1,21 @@
+// Package source provides pluggable variable sources for nomad-pack
```
tgross: You should have copyright headers (with years!) here and in all the new files you've added. You probably don't need this docstring.

internal/pkg/variable/source/base_source_test.go
```diff
@@ -0,0 +1,112 @@
...
+
+func TestBaseSource_Name(t *testing.T) {
+	source := NewBaseSource("custom-name", 10, nil)
+	require.Equal(t, "custom-name", source.Name())
+}
+
+func TestBaseSource_Priority(t *testing.T) {
+	source := NewBaseSource("test", 42, nil)
+	require.Equal(t, 42, source.Priority())
+}
```
tgross: These tests are testing what? Constructors that do nothing but return an object? Aren't all these behaviors already being exercised by tests of the registry (and if not, shouldn't they be)?

internal/pkg/variable/source/base_source_test.go
```diff
@@ -0,0 +1,112 @@
...
+
+func TestBaseSource_Fetch_NilVars(t *testing.T) {
+	source := NewBaseSource("test", 10, nil)
+	ctx := context.Background()
```
tgross: Just in general: https://pkg.go.dev/testing#T.Context

internal/pkg/variable/source/registry.go
```diff
@@ -0,0 +1,110 @@
...
+func NewRegistry() *Registry {
+	return &Registry{
+		sources: make([]VariableSource, 0),
+	}
+}
+
+// Register adds a source to the registry. Returns an error if the source
+// is nil or if a source with the same name is already registered.
+func (r *Registry) Register(source VariableSource) error {
+	if source == nil {
```
tgross: It's not safe to nil-check an interface like this: https://go.dev/play/p/se_S1xh0dj5 See https://rednafi.com/go/nil-interface-comparison/ for a good in-depth explainer.

But also the variable sources aren't third-party plugins; they're code we're writing and adding to the project, right? That makes this is a _programmer's error_ if we had a nil source, so crashing on a nil pointer is not just ok but desirable!

internal/pkg/variable/source/registry.go
```diff
@@ -0,0 +1,110 @@
...
+// Sources returns a copy of all registered sources for inspection.
+// This is useful for debugging and testing.
+func (r *Registry) Sources() []VariableSource {
+	return append([]VariableSource(nil), r.sources...)
+}
+
+// Clear removes all registered sources from the registry.
+func (r *Registry) Clear() {
+	r.sources = r.sources[:0]
+}
```
tgross: Why would we ever clear the registry? This is a CLI tool, so I'd expect the registry to be used and then abandoned when the command exits, right?

internal/pkg/variable/source/registry.go
```diff
@@ -0,0 +1,110 @@
...
+		result = append(result, v)
+	}
+
+	return result, nil
+}
+
+// Sources returns a copy of all registered sources for inspection.
+// This is useful for debugging and testing.
+func (r *Registry) Sources() []VariableSource {
+	return append([]VariableSource(nil), r.sources...)
```
tgross:
```suggestion
	return slices.Clone(r.sources)
```

Although I'd also question the usefulness of this; shouldn't tests be able to look into the private field?

internal/pkg/variable/source/registry.go
```diff
@@ -0,0 +1,110 @@
...
+// Returns an error if context is cancelled or if any source fails to fetch.
+func (r *Registry) Resolve(ctx context.Context, packID pack.ID) ([]*variables.Variable, error) {
+	// Check context before starting
+	if err := ctx.Err(); err != nil {
+		return nil, fmt.Errorf("context cancelled before resolution: %w", err)
+	}
+
+	// Sort by priority (lower first, so higher priority overwrites)
+	sorted := make([]VariableSource, len(r.sources))
+	copy(sorted, r.sources)
```
tgross:
```suggestion
	sorted := slices.Clone(r.sources)
```

But do we need to copy this? Even if we're sorting in place, does the registry get accessed concurrently where we care about mutation?

internal/pkg/variable/source/registry_test.go
```diff
@@ -0,0 +1,243 @@
+package source
```
tgross: Rather than using a mock everywhere in this test, couldn't we just use the `BaseSource` directly? It looks like you can give it a map of variables anyways, and then you wouldn't have to maintain a mock.

tgross (changes requested):
This is a good start @TIMMAREDDYDEEKSHITHA-ship-it 

Most of my comments are on the testing strategy. Unfortunately these tests have the smell of LLM-generated tests that "cover" a lot but don't really test anything _meaningful_. Let's take a fresh pass at this with an eye towards what we want the tests to tell us.

internal/pkg/variable/source/source_test.go
DeekshithaTimmareddy: Yes, I will make a note of it and will use shoeing/test for writing new tests

internal/pkg/variable/source/base_source_test.go
DeekshithaTimmareddy: Yes these were just constructor tests which are not required. I removed this file along with registry_test.go and source_test.go which are mock tests.

internal/pkg/variable/source/source.go
DeekshithaTimmareddy: Added!

internal/pkg/variable/source/registry.go
DeekshithaTimmareddy: Yes , this is programmer's error if we had nil sources because these are internal sources we control, so I removed nil check

internal/pkg/variable/source/registry.go
DeekshithaTimmareddy: I added it thinking it would be useful but didn't consider that CLI tools don't need to reuse the registry. Removed it.

internal/pkg/variable/source/registry.go
DeekshithaTimmareddy: Definitely tests should not look into private fields and they should test the public behaviour. Removed it.

internal/pkg/variable/source/registry.go
DeekshithaTimmareddy: I was thinking that sorting can change the slice so I thought I should copy it first and then assign back like what if something else is reading r.sources while I sort. But sorting in place now because it's single-threaded CLI usage, I will remove copy.

internal/pkg/variable/source/registry_test.go
DeekshithaTimmareddy: I removed this test file.

internal/pkg/variable/source/source.go
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) HashiCorp, Inc.
```
tgross: Wrong header as this should be IBM, and still missing years. You can use [`copywrite`](https://github.com/hashicorp/copywrite) to check these. If we don't have that configured for this repo, we should do that (as a separate PR)

internal/pkg/variable/source/base_source.go
```diff
@@ -0,0 +1,76 @@
...
+
+// Priority returns the precedence level (higher = higher priority).
+func (b *BaseSource) Priority() int {
+	return b.priority
+}
+
+// Fetch retrieves variables for the given pack from the wrapped map.
+// Returns an empty slice if the pack is not found or vars is nil.
+func (b *BaseSource) Fetch(ctx context.Context, packID pack.ID) ([]*variables.Variable, error) {
+	// Check context before processing
```
tgross: This is a `i++ // increment i` kind of comment

```suggestion
```

internal/pkg/variable/parser/parser_v2.go
```diff
@@ -101,21 +108,56 @@ func (p *ParserV2) Parse() (*ParsedVariables, hcl.Diagnostics) {
...
+		return nil, diags.Append(&hcl.Diagnostic{
+			Severity: hcl.DiagError,
+			Summary:  "Failed to register CLI source",
+			Detail:   err.Error(),
+		})
+	}
+
+	// Use context with timeout to prevent hanging on external sources
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
```
tgross, line 136: The change you made here is fine, but just as a guideline in the future: if you're going to make a change in a PR that doesn't address a comment, make sure to call it out to the reviewer so it's not missed. I use the "viewed" button in github to help catch these, but it's very easy for reviewers to miss subtle changes among a pile of other changes.

internal/pkg/variable/source/registry.go
```diff
@@ -0,0 +1,107 @@
...
+		for _, v := range vars {
+			varMap[v.Name] = v
+		}
+	}
+
+	// Convert map back to slice
+	result := make([]*variables.Variable, 0, len(varMap))
+	for _, v := range varMap {
+		result = append(result, v)
+	}
```
tgross:
```suggestion
	result := slices.Collect(maps.Values(varMap))
```

tgross (changes requested):
I've left some more comments. But more importantly in the process of trying to eliminate redundant or meaningless tests, you dropped _all_ the tests.

DeekshithaTimmareddy:
> I've left some more comments. But more importantly in the process of trying to eliminate redundant or meaningless tests, you dropped _all_ the tests.

I wanted to make sure the tests which are include here should be meaningful and not mock tests anymore, So added integration tests, one which verifies that higher priority sources correctly override lower priority ones, empty registry and Context cancellation.

So these tests use real BaseSource instances and test complete Register.

internal/pkg/variable/source/source.go
DeekshithaTimmareddy: Changed!

internal/pkg/variable/parser/parser_v2.go
DeekshithaTimmareddy: Definitely I will make a note of it

internal/pkg/variable/source/registry.go
DeekshithaTimmareddy: Thanks @tgross , I applied this change and it is much cleaner with slices.Collect

internal/pkg/variable/source/source.go
hashi-divyansh: It's already configured for the nomad-pack repository. You @TIMMAREDDYDEEKSHITHA-ship-it  have to run `copywrite headers ` from the root of the repo.

DeekshithaTimmareddy: Yes I noticed and did change!

tgross (approved):
LGTM!

PR #896, opened 2026-06-03
[Issue #473 - Subtask 2] Implement External Variable Sources (Consul KV, Vault KV, Nomad Variables)

DeekshithaTimmareddy:
**Description**

This PR implements Subtask 2 of Issue #473, adding three external variable sources to nomad-pack's pluggable variable architecture.

# Testing

image

image

image

**Reminders**

### DIFF
--- a/internal/pkg/variable/source/consul_source.go
+++ b/internal/pkg/variable/source/consul_source.go
@@ -13,7 +13,6 @@ import (
 	"github.com/hashicorp/nomad-pack/sdk/pack"
 	"github.com/hashicorp/nomad-pack/sdk/pack/variables"
 	"github.com/zclconf/go-cty/cty"
-	"github.com/zclconf/go-cty/cty/gocty"
 )
 
 // ConsulSource fetches variables from Consul KV store.
@@ -120,69 +119,4 @@ func (c *ConsulSource) convertValue(data []byte) (cty.Value, error) {
 
 	// Convert JSON to cty.Value
 	return convertJSONToCty(v)
-}
-
-// convertJSONToCty converts a Go interface{} (from JSON) to a cty.Value.
-func convertJSONToCty(v interface{}) (cty.Value, error) {
-	switch val := v.(type) {
-	case nil:
-		return cty.NullVal(cty.DynamicPseudoType), nil
-
-	case string:
-		return cty.StringVal(val), nil
-
-	case float64:
-		return cty.NumberFloatVal(val), nil
-
-	case bool:
-		return cty.BoolVal(val), nil
-
-	case []interface{}:
-		if len(val) == 0 {
-			return cty.ListValEmpty(cty.DynamicPseudoType), nil
-		}
-
-		// Convert each element
-		elements := make([]cty.Value, len(val))
-		for i, elem := range val {
-			elemVal, err := convertJSONToCty(elem)
-			if err != nil {
-				return cty.NilVal, fmt.Errorf("failed to convert list element %d: %w", i, err)
-			}
-			elements[i] = elemVal
-		}
-
-		// Try to create a list with a unified type
-		return cty.ListVal(elements), nil
-
-	case map[string]interface{}:
-		if len(val) == 0 {
-			return cty.EmptyObjectVal, nil
-		}
-
-		// Convert each value
-		attrs := make(map[string]cty.Value)
-		for k, v := range val {
-			attrVal, err := convertJSONToCty(v)
-			if err != nil {
-				return cty.NilVal, fmt.Errorf("failed to convert object attribute %s: %w", k, err)
-			}
-			attrs[k] = attrVal
-		}
-
-		return cty.ObjectVal(attrs), nil
-
-	default:
-		ty, err := gocty.ImpliedType(v)
-		if err != nil {
-			return cty.NilVal, fmt.Errorf("unsupported type %T: %w", v, err)
-		}
-
-		ctyVal, err := gocty.ToCtyValue(v, ty)
-		if err != nil {
-			return cty.NilVal, fmt.Errorf("failed to convert %T to cty.Value: %w", v, err)
-		}
-
-		return ctyVal, nil
-	}
 }
--- a/internal/pkg/variable/source/consul_source.go
+++ b/internal/pkg/variable/source/consul_source.go
@@ -39,10 +39,6 @@ func NewConsulSource(priority int, config *api.Config, prefix string) (*ConsulSo
 		return nil, fmt.Errorf("failed to create Consul client: %w", err)
 	}
 
-	if prefix == "" {
-		prefix = "nomad-pack/vars"
-	}
-
 	return &ConsulSource{
 		name:     "consul",
 		priority: priority,
@@ -177,7 +173,6 @@ func convertJSONToCty(v interface{}) (cty.Value, error) {
 		return cty.ObjectVal(attrs), nil
 
 	default:
-		// Fallback: try to use gocty to convert
 		ty, err := gocty.ImpliedType(v)
 		if err != nil {
 			return cty.NilVal, fmt.Errorf("unsupported type %T: %w", v, err)
--- a/internal/pkg/variable/source/consul_source.go
+++ b/internal/pkg/variable/source/consul_source.go
@@ -1,0 +1,193 @@
+// Copyright IBM Corp. 2023, 2026
+// SPDX-License-Identifier: MPL-2.0
+
+package source
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/consul/api"
+	"github.com/hashicorp/nomad-pack/sdk/pack"
+	"github.com/hashicorp/nomad-pack/sdk/pack/variables"
+	"github.com/zclconf/go-cty/cty"
+	"github.com/zclconf/go-cty/cty/gocty"
+)
+
+// ConsulSource fetches variables from Consul KV store.
+// Variables are stored under a configurable prefix with the structure:
+// {prefix}/{pack-id}/{variable-name}
+type ConsulSource struct {
+	name     string
+	priority int
+	client   *api.Client
+	prefix   string
+}
+
+// NewConsulSource creates a new Consul KV variable source.
+// The config parameter can be nil to use default Consul configuration
+// (which reads from CONSUL_HTTP_ADDR and CONSUL_HTTP_TOKEN env vars).
+func NewConsulSource(priority int, config *api.Config, prefix string) (*ConsulSource, error) {
+	if config == nil {
+		config = api.DefaultConfig()
+	}
+
+	client, err := api.NewClient(config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create Consul client: %w", err)
+	}
+
+	if prefix == "" {
+		prefix = "nomad-pack/vars"
+	}
+
+	return &ConsulSource{
+		name:     "consul",
+		priority: priority,
+		client:   client,
+		prefix:   strings.TrimSuffix(prefix, "/"),
+	}, nil
+}
+
+// Name returns the unique identifier for this source.
+func (c *ConsulSource) Name() string {
+	return c.name
+}
+
+// Priority returns the precedence level (higher = higher priority).
+func (c *ConsulSource) Priority() int {
+	return c.priority
+}
+
+// Fetch retrieves variables for the given pack from Consul KV.
+// Variables are expected to be stored as JSON values that can be
+// converted to cty.Value types. If a value is not valid JSON,
+// it will be treated as a string.
+func (c *ConsulSource) Fetch(ctx context.Context, packID pack.ID) ([]*variables.Variable, error) {
+	if err := ctx.Err(); err != nil {
+		return nil, err
+	}
+
+	// Build KV path: prefix/packID/
+	path := fmt.Sprintf("%s/%s/", c.prefix, packID)
+
+	// List all keys under this path
+	opts := api.QueryOptions{RequireConsistent: true}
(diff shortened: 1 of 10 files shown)


internal/pkg/variable/source/convert.go
```diff
@@ -0,0 +1,78 @@
...
+	"fmt"
+
+	"github.com/zclconf/go-cty/cty"
+	"github.com/zclconf/go-cty/cty/gocty"
+)
+
+// convertJSONToCty converts a Go interface{} (from JSON) to a cty.Value.
+// This is shared by multiple source implementations that need to convert
+// JSON data to HCL types.
+func convertJSONToCty(v interface{}) (cty.Value, error) {
```
tgross: Modernize by using `any` anywhere you've got `interface{}`. It's an alias so it's exactly the same thing in function but it's _much_ nicer to read.

```suggestion
func convertJSONToCty(v any) (cty.Value, error) {
```

internal/pkg/variable/source/convert.go
```diff
@@ -0,0 +1,78 @@
...
+package source
+
+import (
+	"fmt"
+
+	"github.com/zclconf/go-cty/cty"
+	"github.com/zclconf/go-cty/cty/gocty"
+)
+
+// convertJSONToCty converts a Go interface{} (from JSON) to a cty.Value.
```
tgross: There wasn't any explanation of the design we've introduced here in the PR description or the original issue, and this feature has a good bit of surface area. Let's walk through how I'd expect this to work:

* All 3 APIs return _only_ strings as values (is this actually correct? what about base-64 encoded Vault values?)
* But Pack variables are HCL2 variables, so they can be arbitrarily typed.
* So we're doing a type conversion here under the assumption that the Pack author wants to convert whatever string value they've encoded to a non-string type.

If that were correct then it _should_ be safe because the Pack author is requesting a specific type coercsion. 

But that doesn't seem like it's what we're doing here! Instead, we're taking the value we get from the JSON response body and then _assuming_ we know what type to convert it to based on the shape from the caller? I'm not sure I understand why we'd do that

internal/pkg/variable/source/consul_source.go
```diff
@@ -0,0 +1,122 @@
...
+
+	"github.com/hashicorp/consul/api"
+	"github.com/hashicorp/nomad-pack/sdk/pack"
+	"github.com/hashicorp/nomad-pack/sdk/pack/variables"
+	"github.com/zclconf/go-cty/cty"
+)
+
+// ConsulSource fetches variables from Consul KV store.
+// Variables are stored under a configurable prefix with the structure:
+// {prefix}/{pack-id}/{variable-name}
```
tgross, line 20: Where's this prefix value actually come from? Have we verified that Pack IDs compatible with Consul, Nomad, and Vault KV paths? (I know Nomad jobs can have pretty wild IDs with unicode and slashes and all kinds of stuff we regret... I don't recall what that looks like for Pack)

internal/pkg/variable/source/consul_source_test.go
```diff
@@ -0,0 +1,185 @@
...
+	for key, value := range testData {
+		_, err := kv.Put(&api.KVPair{
+			Key:   key,
+			Value: []byte(value),
+		}, nil)
+		must.NoError(t, err)
+	}
+
+	// Cleanup
+	defer func() {
```
tgross: `t.Cleanup()`

internal/pkg/variable/source/consul_source_test.go
```diff
@@ -0,0 +1,185 @@
...
+	config := api.DefaultConfig()
+	client, err := api.NewClient(config)
+	if err != nil {
+		t.Skipf("Consul client creation failed: %v", err)
+	}
+
+	// Verify Consul is reachable
+	_, err = client.Status().Leader()
+	if err != nil {
+		t.Skipf("Consul not reachable: %v", err)
```
tgross: I'd fail rather than skip here. If you skip, you end up just skipping the tests and silently getting green CI if the test environment is broken.

tgross (commented):
This is a good start, although I've left some design questions I'm puzzling over. Most of the comments I've made on the Consul provider apply to the other two as well.

Also, for future reference this would have been a good PR to break up into chunks. 1100+ lines is a _lot_ to review, and you'd have been able to get one PR in to work thru design and then the other two would be all-but-rubberstamped.

internal/pkg/variable/source/convert.go
DeekshithaTimmareddy: Let me clarify the design and why i chose this approach.

Firstly, to address your question regarding APIs return:
- Consul KV : returns JSON or plain text
- Vault KV : returns strings, maps or structured data
- Nomad Variables : Returns always strings

and regarding base64-encoded Vault values- Vault's API client automatically decodes base64 when reading secrets, so we receive the decoded value.

I understand that you expected something like explicit type coercion, while that would be safe, I believe JSON inference is actually better.

Regarding the current design :
read data  from the external sources, attempt JSON parsing and then JSON parsing will provide type information automatically and if JSON parsing fails then fallback to string.

We are not assuming types based on shapes, so the JSON PARSER  tells us the types. We're simply preserving those types when converting to HCL. What do you say??

internal/pkg/variable/source/consul_source.go
DeekshithaTimmareddy: Here, the prefix comes from the user when they create a source and the prefix is configurable so users can organize their variables under different paths.

And variable source implementatios don't perform their own validation - they accept whatever Pack ID string is passed to them and construct paths accordingly. (After testing) all three sources are fully compatible with valid pack IDs

internal/pkg/variable/source/convert.go
tgross: > Consul KV: returns JSON or plain text

It returns _encoded_ values:

```
$ consul kv put nomad/name "Tim"

$ curl -sH "X-Consul-Token: $CONSUL_HTTP_TOKEN" "http://localhost:8500/v1/kv/nomad/name" | jq .
[
  {
    "LockIndex": 0,
    "Key": "nomad/name",
    "Flags": 0,
    "Value": "VGlt",
    "Partition": "default",
    "Namespace": "default",
    "CreateIndex": 45,
    "ModifyIndex": 45
  }
]

$ echo "VGlt" | base64 -d
Tim%

$ consul kv put "something-else" '{"foo": "bar"}'
Success! Data written to: something-else

$ curl -sH "X-Consul-Token: $CONSUL_HTTP_TOKEN" "http://localhost:8500/v1/kv/something-else" | jq .
[
  {
    "LockIndex": 0,
    "Key": "something-else",
    "Flags": 0,
    "Value": "eyJmb28iOiAiYmFyIn0=",
    "Partition": "default",
    "Namespace": "default",
    "CreateIndex": 322,
    "ModifyIndex": 322
  }
]

$ curl -sH "X-Consul-Token: $CONSUL_HTTP_TOKEN" "http://localhost:8500/v1/kv/something-else" | jq .

$ echo "eyJmb28iOiAiYmFyIn0=" | base64 -d | jq .
{
  "foo": "bar"
}
```

> Vault KV : returns strings, maps or structured data

Vault has it's own structure for the value:

```
$ vault kv put -mount secret something name=Tim

$ curl -sH "X-Vault-Token: $VAULT_TOKEN" "http://localhost:8200/v1/secret/data/something" | jq .
{
  "request_id": "f42ee269-3990-ccdd-e037-75054ce1770d",
  "lease_id": "",
  "renewable": false,
  "lease_duration": 0,
  "data": {
    "data": {
      "name": "Tim"
    },
    "metadata": {
      "created_time": "2026-06-09T14:39:52.677197287Z",
      "custom_metadata": null,
      "deletion_time": "",
      "destroyed": false,
      "version": 1
    }
  },
  "wrap_info": null,
  "warnings": null,
  "auth": null,
  "mount_type": "kv"
}
```

But if you try to give it structured data, it's just an encoded string:

```
$ vault kv put -mount secret something name='{"foo": "bar"}'
==== Secret Path ====
secret/data/something

======= Metadata =======
Key                Value
---                -----
created_time       2026-06-09T14:43:05.283026804Z
custom_metadata    <nil>
deletion_time      n/a
destroyed          false
version            2

$ curl -sH "X-Vault-Token: $VAULT_TOKEN" "http://localhost:8200/v1/secret/data/something" | jq .
{
  "request_id": "ef41917c-0657-6e1d-d4a7-ea8b3d97b21c",
  "lease_id": "",
  "renewable": false,
  "lease_duration": 0,
  "data": {
    "data": {
      "name": "{\"foo\": \"bar\"}"
    },
    "metadata": {
      "created_time": "2026-06-09T14:43:05.283026804Z",
      "custom_metadata": null,
      "deletion_time": "",
      "destroyed": false,
      "version": 2
    }
  },
  "wrap_info": null,
  "warnings": null,
  "auth": null,
  "mount_type": "kv"
}

```

So in other words, while the API response is structured the _value_ returned for a given item is always a string! It's just that the string may be encoded data. But from the perspective of Pack you have no idea whether that encoded string is safe to decode -- it could be binary-encoded data for all you know! The only way to know whether it's safe to decode to a particular format is if there's a schema that says so.

internal/pkg/variable/source/consul_source.go
tgross: > the prefix comes from the user 

Sure, but how? There's no design doc here so there's no indication to me how we intended to configure these sources.

internal/pkg/variable/source/convert.go
DeekshithaTimmareddy: You are right!! Here we are guessing types instead of being told explicitly. So do you want me to document our convention and also add explicit type information if users need it?

internal/pkg/variable/source/consul_source.go
DeekshithaTimmareddy: OK, I will document explaining the configuration and usage.

internal/pkg/variable/source/consul_source_test.go
DeekshithaTimmareddy: I've updated tests to use t.Fatalf() as you suggested. But CI is now failing because Github Actions workflow doesn't have Consul and Vault services configured. So should i add Consul and Vault service containers to CI workflow??

tgross: Seems reasonable.

internal/pkg/variable/source/convert.go
tgross: > and also add explicit type information if users need it?

I'm not sure I understand what you're proposing here. Remember that our goal is to get data from these sources into Pack variable values. And those variables already have type information in their schema (in the jobspec). The problem I'm trying to avoid is one similar to https://github.com/hashicorp/nomad/pull/28095. 

Let's walk through an example. Suppose I have a Consul variable with the value `eyJmb28iOiAiYmFyIn0=` (base64 encoded `{"foo": "bar"}`). Also suppose I have Pack variable definition like this:

```hcl
variable "task_env" {
  type = string
}
```

What value is the user expecting to see when they render this? Let's look at some possibilities:

```hcl
job "example" {
  # user is expecting it to be decoded for them
  # (this seems like a terrible option)
  env = [[ .packname.task_env ]]
}
```

vs

```hcl
job "example" {
  env = [[ .packname.task_env | fromJson ]]
}
```

vs 

```hcl
job "example" {
  env = {
     # oops! this is getting decoded somewhere in the workload itself!
     foo = [[ .pack.name.task_env ]]
  }
}
```

Even if we document a particular way we expect the data to be decoded, any choice we make here is going to constrain the user who just put a string in the KV store and expects to _exactly that data_ back. Trying to guess how they want to consume it seems like it's going to inevitably break for someone, somewhere.

DeekshithaTimmareddy: Looking at issue, I understand the problem with automatic JSON parsing, i think it can break user expectations when they store JSON strings that should remain as strings. And also pack's variables.hcl already has type information that should guide the conversion.

So instead of sources doing type conversion, they should return raw string values and let variable decoder use pack's schema to determine how to parse. Do you want me to refactor the code to implement this?? If so, i will:
- Remove JSON parsing from source implementations
- Move current logic to variable decoder
- And also add schema-aware parsing

tgross: That sounds _roughly_ correct but what does "schema-aware parsing" actually mean here?

DeekshithaTimmareddy: It means using the pack's variable type defs to determine how to convert string value instead of guessing based on whether it looks like JSON.

tgross: I think you need to start a feature branch and start working through this more methodically. Starting with a PR of the source readers with no consumers makes it hard to have a useful conversation about how the values are making their way to the values. I have no idea where you intend on adding that code, but it's only going to further inflate an already huge PR.

DeekshithaTimmareddy: ok so in this pr i will focus only on source implementations and next i will be doing CLI integration in seperate pr so that it will be easy to discuss schema-aware parsing in further pr

tgross: We shouldn't be landing large PRs like this in main without understanding how they're being consumed. So here's what you should do:

* Start a new feature branch from main
* Rebase this PR on that branch and then edit the target of this PR to point to that branch
* Open a new PR, also targeting that feature branch, addressing consumer side

hashi-divyansh:
@DeekshithaTimmareddy My recommendation would be to stop working on this PR and divide your ticket into multiple subtasks. The logical code changes should not be more than 100 lines; otherwise, it will be really difficult to review.
We will review and merge your subtasks' PRs into `feature/external-variable-sources`, and after integration testing, merge `feature/external-variable-sources` into the `main` branch."

DeekshithaTimmareddy:
Closing in favour of #898

PR #898, opened 2026-06-11
[Issue #473] Add Consul KV variable source 

DeekshithaTimmareddy:
**Description**

This PR implements Subtask 2a of Issue #473 (External Variable Sources for
nomad-pack). It adds support for fetching pack variables from Consul KV, a shared
JSON→cty conversion utility, and the `--var-source` CLI flag with full
integration.

### Key behaviors
- **URL-based config** following standard URL rules:
  - `consul:///<prefix>` — Consul address from the environment (e.g. `CONSUL_HTTP_ADDR`)
  - `consul://<host>:<port>/<prefix>` — explicit Consul address
 
- **Default KV layout:** `<prefix>/<pack-name>/<variable-name>`.
  Pass `?full-path=true` to use `<prefix>` as-is (pack name not appended).

- **Type-aware conversion:** a Pack variable typed `string` keeps the raw Consul
  bytes; only non-string types are JSON-decoded — no type-guessing before the
  schema is known.

- **Parse/build split:** only `run`, `plan`, and `render` read remote sources.
  `stop` and `generate varfile` do **not** accept `--var-source`, so stopping a
  job never depends on a remote source still being reachable.

- **Variable precedence (highest → lowest):**
  1. CLI flags (`--var`)
  2. External sources (`--var-source`)
  3. Variable files (`--var-file`)
  4. Environment variable

## Testing

```bash
# 1. Start Consul
consul agent -dev

# 2. Store variables in Consul KV
#    Default layout is <prefix>/<pack-name>/<variable-name>
consul kv put nomad-pack/my-pack/replicas 3
consul kv put nomad-pack/my-pack/region "us-west-2"

# 3. Render pack with Consul variables  (note the THREE slashes: consul:///)
nomad-pack render my-pack --var-source=consul:///nomad-pack

# 4. Priority: CLI --var overrides Consul
nomad-pack render my-pack --var-source=consul:///nomad-pack --var replicas=5

# 5. Explicit Consul address instead of CONSUL_HTTP_ADDR
nomad-pack render my-pack --var-source=consul://localhost:8500/nomad-pack

# 6. Use the prefix exactly as-is (do NOT append the pack name)
nomad-pack render my-pack --var-source='consul:///my/custom/path?full-path=true'
```

# Test Data

### Store variables in Consul:

image

### Priority Override:

image

image

### Multiple sources:

image

Priority: CLI > Consul File

**Reminders**


tgross (changes requested):
Still no consumer code.

DeekshithaTimmareddy:
> Still no consumer code.

Added! adding integration tests and documentation will make this pr huge so limiting this pr to CLI integration and consumer code only

internal/cli/commands.go
```diff
@@ -267,6 +271,20 @@ func (c *baseCommand) flagSet(bit flagSetBit, f func(*flag.Sets)) *flag.Sets {
...
+			Target:  &c.varSources,
+			Default: make([]string, 0),
+			Usage: `Specifies an external variable source URL. Currently supported:
+					consul:// (Consul KV)
+					
+					Can be specified multiple times. Variables from external sources
+					are merged with file and CLI variables according to precedence rules.
+					
+					Example:
+					  --var-source=consul://nomad-pack`,
```
tgross: This help text doesn't describe what comes after `consul://`. i.e. what does `nomad-pack` refer to here? The help text should also include the critical detail that this requires you have the appropriate environment variables for Consul set, maybe even have a link to their docs page. We also haven't described what the precedence rules are here. They seem short enough that we could do so?

internal/cli/helpers.go
```diff
@@ -40,7 +41,17 @@ func initPackCommand(cfg *caching.PackConfig) (errorContext *errors.UIErrorConte
...
 // generatePackManager is used to generate the pack manager for this Nomad Pack run.
-func generatePackManager(c *baseCommand, client *api.Client, packCfg *caching.PackConfig) *manager.PackManager {
+func generatePackManager(c *baseCommand, client *api.Client, packCfg *caching.PackConfig) (*manager.PackManager, error) {
+	// Parse external variable sources if provided
+	var externalSources []source.VariableSource
+	if len(c.varSources) > 0 {
+		sources, err := parseVarSources(c.varSources, packCfg.Name)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse variable sources: %w", err)
+		}
```
tgross: Contrary to the name "parseVarSources" this function call actually _fetches_ the variables from a remote source. And because this `generatePackManager` is used by several CLI commands, it means we have to fetch all the variables from all the sources every time we run those commands.

While that seems desirable for `nomad-pack run` and `nomad-pack plan`, does it makes sense for `nomad-pack stop` and `nomad-pack generate varfile`? Especially given that you may be stopping a job _because_ its variables no longer exist!

It feels like we probably want to split parsing the variable source configuration from the actual API calls to the remote source here.

internal/cli/varsource_parser.go
```diff
@@ -0,0 +1,124 @@
...
+		// No host specified - this shouldn't happen with consul://prefix format
+		// but handle it gracefully
+		path := strings.Trim(u.Path, "/")
+		if path == "" {
+			return nil, fmt.Errorf("consul URL must include a prefix path (e.g., consul:///nomad-pack)")
+		}
+		prefix = path
+	} else {
+		// Host is specified - could be actual host or the prefix
+		// Check if it looks like a host (has port or is localhost/IP)
```
tgross: > if it looks like a host

_Do not guess!_ Either `url.Parse` parsed it as a hostname or it didn't. You should be doing something more like this: https://go.dev/play/p/dme9yV0owhb that way `consul:///prefix/foo` is the correct way of leaving off the host, and the user intent is unambiguous.

And this is where the CLI help text is missing everything the user would need to use this feature! There's nothing there that even suggests that this URL-shaped string is actually a URL that follows the usual rules and could accept a hostname!

internal/pkg/manager/manager.go
```diff
@@ -66,13 +68,23 @@ func (pm *PackManager) ProcessVariableFiles() (*parser.ParsedVariables, []*error
 	// without the version.
 	parentName, _, _ := strings.Cut(path.Base(pm.cfg.Path), "@")
 
+	// Convert external sources to any slice to avoid import cycle
+	var externalSources []any
```
tgross: If you define the source interface here in the caller package you get type safety without import cycles.

internal/pkg/variable/source/consul_source.go
```diff
@@ -0,0 +1,134 @@
...
+	// Try to parse as JSON first
+	var v any
+	if err := json.Unmarshal(data, &v); err != nil {
+		// If not valid JSON, treat as string
+		return cty.StringVal(string(data)), nil
+	}
+
+	// Convert JSON to cty.Value
+	return convertJSONToCty(v)
+}
```
tgross: Ok, so this just moved the issue from the other PR over here without justifying why we're still doing it. At the time you unmarshal these bodies, you have no idea whether it's safe to do this!
Everything I said in https://github.com/hashicorp/nomad-pack/pull/896#discussion_r3384491514 still applies here.

Why are we _guessing_ what type the user wants with these values before we're anywhere near their schema? If I encode JSON into my Consul KV value but plan on giving it as a string value in my Pack variable (because my workload needs it as a JSON string), this breaks my Pack.

internal/pkg/variable/source/consul_source.go
```diff
@@ -0,0 +1,134 @@
...
+
+	"github.com/hashicorp/consul/api"
+	"github.com/hashicorp/nomad-pack/sdk/pack"
+	"github.com/hashicorp/nomad-pack/sdk/pack/variables"
+	"github.com/zclconf/go-cty/cty"
+)
+
+// ConsulSource fetches variables from Consul KV store.
+// Variables are stored under a configurable prefix with the structure:
+// {prefix}/{pack-id}/{variable-name}
```
tgross: Why have a prefix at all instead of letting the user specify the entire path to the variable name? i.e. `{user defined path}/{variable-name}`. This seems to assume that the user has a tree in Consul KV where all their Pack KVs live like:

```
prefix
├── job1
│   ├── bar
│   ├── baz
│   └── foo
└── job2
    ├── buzz
    └── fizz
```

But why are we assuming that?

internal/pkg/variable/source/consul_source.go
```diff
@@ -0,0 +1,134 @@
...
+// Edge Cases:
+//   - Returns nil (not error) if no keys found at path
+//   - Skips directory entries (keys ending with /)
+//   - Invalid JSON values are treated as strings
+//
+// The parser automatically provides a 30-second timeout context.
+func (c *ConsulSource) Fetch(ctx context.Context, packID pack.ID) ([]*variables.Variable, error) {
+	if err := ctx.Err(); err != nil {
+		return nil, err
+	}
```
tgross: We _just_ checked `ctx.Err()` in the caller (in `Registry.Resolve`), and nothing meaningful is going to happen between then microseconds later when we `List` call below anyways. I'd leave this out.

internal/pkg/variable/parser/parser_v2.go
```diff
@@ -108,22 +115,38 @@ func (p *ParserV2) Parse() (*ParsedVariables, hcl.Diagnostics) {
...
 			Detail:   err.Error(),
 		})
 	}
-	if err := p.sourceRegistry.Register(source.NewCLISource(30, p.flagOverrideVars)); err != nil {
+
+	// Register external sources (Consul, Vault, Nomad) if provided
+	if len(p.cfg.ExternalSources) > 0 {
+		for _, extSrc := range p.cfg.ExternalSources {
+			if src, ok := extSrc.(source.VariableSource); ok {
+				if err := p.sourceRegistry.Register(src); err != nil {
```
tgross: Looking at how this is consumed makes me wonder if `Register` should really be falliable. These aren't external plugins, so an error here is a programmer error, not a runtime error, isn't it?

tgross (changes requested):

internal/cli/commands.go
DeekshithaTimmareddy: yes , I added the link to the docs and also updated the help text!!

internal/cli/helpers.go
DeekshithaTimmareddy: I have not thought about this, it doesn't makes sense for registering on stop also. I updated this "--var-source" which registers on run/plan/render , so stop and generate don't accept it at all

internal/cli/varsource_parser.go
DeekshithaTimmareddy: Thanks Tim, i changed it. Now, the code uses this standard URL parsing and there is no guessing

internal/pkg/manager/manager.go
DeekshithaTimmareddy: I think this interface in manager was actually not used, so i redid it with interface as you mentioned in source package and now it is fully typed and also i removed this dead interface and map also.

internal/pkg/variable/source/consul_source.go
DeekshithaTimmareddy: I got it, i think it can break user expectations when they store JSON strings that should remain as strings. And also pack's variables.hcl already has type information that should guide the conversion.

So now the source uses variable's declared type : if it's string it returns raw bytes as it is, only non-string types get JSON parsed

internal/pkg/variable/source/consul_source.go
DeekshithaTimmareddy: I mean here i thought this could be the default structure for convenience but thinking aboutt this full path using makes sense. So i was thinking that why not both use cases like default and custom paths, i implemented this, let me know your opinion

internal/pkg/variable/source/consul_source.go
DeekshithaTimmareddy: Yes, i removed it

internal/pkg/variable/parser/parser_v2.go
DeekshithaTimmareddy: Yes, Register errors are programmer errors here, so i've designed Register to not return errors at all. Source name uniqueness is enforced by source constructors here, So this makes duplicate registration impossible by design. What do you say??

internal/pkg/variable/source/consul_source.go
tgross: What convenience? What's the workflow you're envisioning here? Step back away from the code and look at what you're asking the user to do which is to specifiy _two_ values for the path and only one of them is explicit and the other is being constructed implicitly from the pack name. You haven't bothered to justify this

internal/cli/commands.go
```diff
@@ -267,6 +271,76 @@ func (c *baseCommand) flagSet(bit flagSetBit, f func(*flag.Sets)) *flag.Sets {
...
+				Name:    "var-source",
+				Target:  &c.varSources,
+				Default: make([]string, 0),
+				Usage: `Specifies an external variable source URL. Currently supported:
+					  • Consul KV - Fetch variables from Consul
+					
+					The value is a standard URL and follows the usual URL rules,
+					including an optional host. Use the consul:// scheme:
+
+					  consul://<host>:<port>/<prefix>[?options]
```
tgross: Get rid of the prefix. Require a path.

internal/cli/commands.go
```diff
@@ -267,6 +271,76 @@ func (c *baseCommand) flagSet(bit flagSetBit, f func(*flag.Sets)) *flag.Sets {
...
+					  consul kv put config/webapp/region "us-west-2"
+					  
+					  # 2. Run pack with Consul variables (uses CONSUL_HTTP_ADDR from env)
+					  nomad-pack run webapp --var-source=consul:///config
+					  
+					  # 3. Or specify Consul address explicitly
+					  nomad-pack run webapp --var-source=consul://localhost:8500/config
+					  
+					  # 4. Override specific variables via CLI (highest precedence)
+					  nomad-pack run webapp --var-source=consul:///config --var replicas=5`,
```
tgross: Our terminal help text should not read like LLM slop.

internal/cli/varsource_config.go
```diff
@@ -0,0 +1,128 @@
...
+		cfg.Address = u.Host
+	}
+
+	// The path is the KV path
+	path := strings.Trim(u.Path, "/")
+	if path == "" {
+		return cfg, fmt.Errorf("consul URL must include a path (e.g., consul:///nomad-pack)")
+	}
+
+	cfg.Path = path
```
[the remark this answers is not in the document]
tgross: Fix this, but ignore the poor suggestion they've given here.

internal/cli/commands.go
```diff
@@ -267,6 +271,76 @@ func (c *baseCommand) flagSet(bit flagSetBit, f func(*flag.Sets)) *flag.Sets {
...
+
+					By default the pack's name (or alias) is appended to the prefix,
+					so each variable is read from:
+
+					  <prefix>/<pack-name>/<variable-name>
+
+					Query options (quote the URL in your shell when using them):
+					  full-path=true  Use <prefix> verbatim; do NOT append the pack
+					                  name. Variables are read from
+					                  <prefix>/<variable-name>.
```
tgross: Why is this configurable? This is a mess of knobs you absolutely do not need. Just tell the user to give you a path to where the variables live. Why are we making this so complicated?

internal/cli/commands.go
```diff
@@ -267,6 +271,76 @@ func (c *baseCommand) flagSet(bit flagSetBit, f func(*flag.Sets)) *flag.Sets {
...
+					  token=<token>   Consul ACL token. Prefer CONSUL_HTTP_TOKEN so the
+					                  token is not leaked into your shell history.
```
tgross: I'd strongly recommend against even giving this option.

internal/pkg/variable/source/config.go
```diff
@@ -0,0 +1,59 @@
...
+type SourceConfig interface {
+	// Build constructs the concrete VariableSource described by this config.
+	// Implementations may create API clients here, but must not perform
+	// remote reads; reads happen in VariableSource.Fetch.
+	Build() (VariableSource, error)
+}
+
+// ConsulSourceConfig holds the parsed configuration for a Consul KV variable
+// source. It is a plain value type with no live connections, making it safe to
+// pass across package boundaries without import cycles or guessing.
```
tgross: Or guessing?

internal/pkg/variable/source/config.go
```diff
@@ -0,0 +1,59 @@
...
+package source
+
+import (
+	"github.com/hashicorp/consul/api"
+)
+
+// SourceConfig is a parsed, lazily-evaluated configuration for a variable
+// source. Parsing a config (for example from a --var-source URL) is kept
+// separate from Build, which constructs the concrete VariableSource and may
+// open remote clients. This separation lets callers parse and validate source
```
tgross: Do we expect `Build` will open clients? That doesn't seem to be the case in our implementation here. We're constructing the `api.Client` but I don't _think_ that makes a connection, does it?

internal/pkg/variable/source/consul_source.go
```diff
@@ -0,0 +1,187 @@
...
+	var v any
+	if err := json.Unmarshal(data, &v); err != nil {
+		return cty.NilVal, fmt.Errorf("expected %s but value is not valid JSON: %w", expectedType.FriendlyName(), err)
+	}
+
+	// Convert JSON to cty.Value
+	ctyVal, err := convertJSONToCty(v)
+	if err != nil {
+		return cty.NilVal, err
+	}
```
tgross: I'm not sure why we're JSON decoding here. Can't HCL decode this bytestring directly?

DeekshithaTimmareddy: Yes, I am not aware of this but after going through, I got to know that cty/json.Marshal does the whole thing in one step, so i switched to that and deleted converter

internal/pkg/variable/source/config.go
DeekshithaTimmareddy: Yes, it doesn't, just builds the client struct and no connection is made until List call in Fetch and i did change in the comment saying this

internal/pkg/variable/source/config.go
DeekshithaTimmareddy: Yeah, "guessing" didn't mean anything there, that was lefover from an earlier draft. I removed it

internal/cli/commands.go
DeekshithaTimmareddy: I removed this option and now it only comes from CONSUL_HTTP_TOKEN

internal/cli/commands.go
DeekshithaTimmareddy: Yes, this is overcomplicated, so i changed it and now you just give a path and variables are read straight from it

internal/cli/commands.go
DeekshithaTimmareddy: Done !

internal/cli/varsource_config.go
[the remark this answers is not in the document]
DeekshithaTimmareddy: ok , I did, now it takes URL by value instead of holding pointer !

internal/pkg/variable/source/consul_source.go
DeekshithaTimmareddy: Yeah, fair point, i have assumed this layout will be convenient without thinking about the workflow. I've dropped that and now the path is using exactly as given, so there's only one explicit value and nothing built from pack.

internal/pkg/variable/source/consul_source.go
```diff
@@ -0,0 +1,171 @@
...
+// SPDX-License-Identifier: MPL-2.0
+
+package source
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/consul/api"
```
tgross, line 11: I checked out this branch and ran `make`, and `go.mod`/`go.sum` needs tidying now because we're importing the Consul API (and there's a fixture file that's not `hclfmt`'d). Let's fix those in this PR so that it's green locally.

I don't know why this doesn't get picked up in CI, but that'd be a good follow-up task.

internal/cli/commands.go
```diff
@@ -267,6 +271,36 @@ func (c *baseCommand) flagSet(bit flagSetBit, f func(*flag.Sets)) *flag.Sets {
...
+					Consul environment configuration (CONSUL_HTTP_ADDR, CONSUL_HTTP_TOKEN,
+					and so on). Each variable is read from <path>/<variable-name>, so
+					include any per-pack grouping in the path yourself.
+
+					Higher-priority sources win on conflict: --var overrides --var-source,
+					which overrides --var-file and the environment.
+
+					Examples:
+					  consul:///config
+					  consul://localhost:8500/config`,
```
tgross: Have you looked at how this usage string actually gets printed? The whitespace gets removed, so having big line breaks as separators doesn't help and it results in some muddled test because it's therefore missing punctuation:

```
$ nomad-pack run --help
...
      --var-source=<string>
        Specifies an external variable source as a URL. May be given
        more than once to read from several sources. Consul KV is the
        only source currently supported, using the consul:// scheme:
        consul://<host>:<port>/<path> The host is optional; omit it
        (consul:///<path>) to use the standard Consul environment
        configuration (CONSUL_HTTP_ADDR, CONSUL_HTTP_TOKEN, and so on).
        Each variable is read from <path>/<variable-name>, so include any
        per-pack grouping in the path yourself. Higher-priority sources
        win on conflict: --var overrides --var-source, which overrides
        --var-file and the environment. Examples: consul:///config
        consul://localhost:8500/config
```

The phrase "overrides -var-file and the environment" leave it ambiguous as to whether `-var-file` overrides the environment or not.

internal/cli/varsource_config.go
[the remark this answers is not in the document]
tgross: My apologies, I should have explained better how this Wiz finding is terrible 😁 The Wiz scan starts off by talking about concurrency but then talks about injection. You've fixed concurrency issues (which were a false positive) but we're still messing with the URL path after the fact instead of changing the inputs to `url.Parse` to be correct in the first place up in `parseVarSourceConfig`. For what it's worth, this is _also_ a false postive in terms of being a security issue (this is input to a command the user is asking us to do!). But it's a good practice to keep the URL parsing all in one place.

internal/pkg/variable/parser/parser_v2.go
```diff
@@ -108,36 +115,36 @@ func (p *ParserV2) Parse() (*ParsedVariables, hcl.Diagnostics) {
...
-			Detail:   err.Error(),
-		})
-	}
-	if err := p.sourceRegistry.Register(source.NewCLISource(30, p.flagOverrideVars)); err != nil {
-		return nil, diags.Append(&hcl.Diagnostic{
-			Severity: hcl.DiagError,
-			Summary:  "Failed to register CLI source",
-			Detail:   err.Error(),
-		})
+	// Register sources with the registry in priority order
```
tgross, line 118: Now that we have some consuming code, I sort of wonder why `sourceRegistry` is an instance member of `ParserV2`? It gets populated in this function, consumed, and then thrown away when we exit (this is a CLI tool, not a service, after all).

Let's leave it for now in this PR but maybe something to come back and think about later.

internal/pkg/variable/source/config.go
```diff
@@ -0,0 +1,50 @@
...
+// SourceConfig is a parsed, lazily-evaluated configuration for a variable
+// source. Parsing a config (for example from a --var-source URL) is kept
+// separate from Build, which constructs the concrete VariableSource. Build does
+// only local work, such as building API client structs; it opens no connection.
+// No network I/O happens until variables are actually fetched for rendering, so
+// callers can parse and validate source configuration up front for free.
+type SourceConfig interface {
+	// Build constructs the concrete VariableSource described by this config.
+	// Implementations may construct API clients here, but constructing a client
+	// does not open a connection; no remote reads happen until
```
tgross:
```suggestion
	// Implementations may construct API clients here, but constructing a client
	// must not open a connection; no remote reads happen until
```

This sounds like a subtle detail, but if you're defining an interface, you're defining a contract for implementing code. This isn't "today constructing a client doesn't open a connection" it's "we _expect_ that constructing a client doesn't open a connection."

That gives future developers a guide to what you were thinking and tells them that they need to change how they've implemented `Build()` if the client constructor behavior were to change.

internal/pkg/variable/source/consul_source.go
```diff
@@ -0,0 +1,171 @@
...
+//   - Returns nil (not error) if no keys found at path
+//   - Skips directory entries (keys ending with /)
+//   - Skips variables not defined in the pack schema
+//
+// The parser wraps Fetch in a timeout context, so a slow or unreachable Consul
+// fails the resolve instead of hanging.
+func (c *ConsulSource) Fetch(ctx context.Context, packID pack.ID, schema map[variables.ID]*variables.Variable) ([]*variables.Variable, error) {
+	// Variables are read from <path>/<variable-name>. Any per-pack namespacing is
+	// part of the path itself, so the packID parameter is intentionally unused here.
+	path := c.path + "/"
```
tgross: Didn't we trim off a `/` when we set up the source?

internal/pkg/variable/source/consul_source.go
```diff
@@ -0,0 +1,171 @@
...
+//   - Variables not in schema are skipped (not returned)
+//
+// Edge Cases:
+//   - Returns nil (not error) if no keys found at path
+//   - Skips directory entries (keys ending with /)
+//   - Skips variables not defined in the pack schema
+//
+// The parser wraps Fetch in a timeout context, so a slow or unreachable Consul
+// fails the resolve instead of hanging.
+func (c *ConsulSource) Fetch(ctx context.Context, packID pack.ID, schema map[variables.ID]*variables.Variable) ([]*variables.Variable, error) {
```
tgross:
```suggestion
func (c *ConsulSource) Fetch(ctx context.Context, _ pack.ID, schema map[variables.ID]*variables.Variable) ([]*variables.Variable, error) {
```

This suggests we may just not need this field in the interface, but fine to keep it for now and see if we really need it in the other sources.

internal/pkg/variable/source/consul_source.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	vars := make([]*variables.Variable, 0, len(pairs))
+	for _, pair := range pairs {
+		varName := strings.TrimPrefix(pair.Key, path)
+
+		// Skip if this is a directory (ends with /)
+		if strings.HasSuffix(varName, "/") {
+			continue
+		}
```
tgross: Nitpick: you can do this `HasSuffix` check before you do the `TrimPrefix` mutation

internal/pkg/variable/source/consul_source.go
```diff
@@ -0,0 +1,171 @@
...
+
+		// An empty Consul value can't be decoded into a non-string type. Skip it
+		// so the variable falls back to its default instead of failing the entire
+		// resolve (which would also drop the other variables from this source).
+		// Empty values for string variables are intentionally kept as "".
+		if len(pair.Value) == 0 && schemaVar.Type != cty.String {
+			continue
+		}
```
tgross: Why not treat this as an error, so that the user knows there's something goofy going on with the values in Consul?

tgross:
Tested end-to-end:

```
$ nomad-pack generate pack example
...

$ consul kv put pack/count 3
Success! Data written to: pack/count

$ consul kv put pack/datacenters '["dc1"]'
Success! Data written to: pack/datacenters

$ consul kv put pack/message Hi
Success! Data written to: pack/message

$ consul kv put pack/region philly
Success! Data written to: pack/region

$ consul kv put pack/register_service true
Success! Data written to: pack/register_service

$ consul kv put pack/service_name "www"
Success! Data written to: pack/service_name

$ consul kv put pack/service_tags '["whatever"]'
Success! Data written to: pack/service_tags

$ nomad-pack render --var-source consul:///pack example
example/example.nomad:

job "example" {
  region      = "philly"
  datacenters = ["dc1"]
  type        = "service"

  group "app" {
    count = 3

    network {
      port "http" {
        to = 8000
      }
    }


    service {
      name     = "www"
      tags     = ["whatever"]
      provider = "nomad"
      port     = "http"
      check {
        name     = "alive"
        type     = "http"
        path     = "/"
        interval = "10s"
        timeout  = "2s"
      }
    }


    restart {
      attempts = 2
      interval = "30m"
      delay    = "15s"
      mode     = "fail"
    }

    task "server" {
      driver = "docker"

      config {
        image = "mnomitch/hello_world_server"
        ports = ["http"]
      }

      env {
        MESSAGE = "Hi"
      }
    }
  }
}

```

tgross (commented):
This is looking pretty good! A few more minor items to resolve and I think this will be ready to merge.

internal/pkg/variable/source/consul_source.go
DeekshithaTimmareddy: I think there's no meaningful "empty" number or bool so an empty value there almost always means Consul key is misconfigured, So i switched it to an error!

internal/pkg/variable/source/consul_source.go
DeekshithaTimmareddy: Moved "HasSuffix" directory check above "TrimPrefix"

internal/pkg/variable/source/consul_source.go
DeekshithaTimmareddy: Done, I dropped it to _pack.ID. Once other sources land we can see whether this earns its place or not

internal/pkg/variable/source/consul_source.go
DeekshithaTimmareddy: Yes , we trim slashes when building the source so c.path is stored clean and I add single "/" here for two reasons : it scopes KV "List" to keys under path ( so let say for eg: demoX doesn't also match demoXy) and it gives "TrimPrefix" right prefix to strip each key. So i added a comment so it's not confusing.

internal/pkg/variable/parser/parser_v2.go
DeekshithaTimmareddy: Yeah, it doesn't really need to be a struct field but I'll make a note to look into it

internal/cli/commands.go
DeekshithaTimmareddy: I hadn't looked at collapsed output closely enough. Now i rewrote it and made precedence as explicit ordered list. Here's how it renders now:

image

internal/pkg/variable/source/consul_source.go
DeekshithaTimmareddy: I fix it first but i thought this is not related to this issue and i reverted it back, so now i ran hclfmt on fixture and it just realigned "=" in items block and i also agree about this CI gap which is worth a follow-up because it's not running hclfmt on fixtures path.

DeekshithaTimmareddy:
> This is looking pretty good! A few more minor items to resolve and I think this will be ready to merge.

After this pr got merged, i will be focusing on integration tests and also documentation of this consul source in the follow up pr, let me know your comments

tgross (approved):
:shipit:

PR #905, opened 2026-06-25
[Issue #473] Add Nomad Variables variable source

DeekshithaTimmareddy:
**Description**

This PR implements Subtask 4 of Issue https://github.com/hashicorp/nomad-pack/issues/473 (External Variable Sources for nomad-pack). It adds support for fetching pack variables from Nomad Variables along with the integration tests coverage

E2E Testing:

image

image

image

**Reminders**

### DIFF
--- a/internal/pkg/variable/source/config.go
+++ b/internal/pkg/variable/source/config.go
@@ -103,12 +103,21 @@ type NomadSourceConfig struct {
 }
 
 // Build implements SourceConfig by constructing a NomadSource.
+//
+// When no host is supplied (the nomad:///path form), the address and every
+// other client setting come entirely from the standard Nomad environment via
+// nomadapi.DefaultConfig() — NOMAD_ADDR, NOMAD_TOKEN, NOMAD_NAMESPACE, and so
+// on flow through unchanged. NOMAD_ADDR may point at a unix socket (unix://…),
+// so running Pack inside a Nomad task and aiming it at the task's API socket
+// needs no special handling here: the nomad/api client dials the socket
+// natively and the workload's token authenticates as usual.
 func (n NomadSourceConfig) Build() (VariableSource, error) {
 	apiCfg := nomadapi.DefaultConfig()
 	if n.Address != "" {
-		// Nomad's API address must be a full URL (unlike Consul's host:port), so
-		// default to http when the URL omits a scheme — matching Nomad's own
-		// default address (http://127.0.0.1:4646).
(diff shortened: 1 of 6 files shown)


internal/pkg/variable/source/config.go
```diff
@@ -85,3 +86,35 @@ func (v VaultSourceConfig) Build() (VariableSource, error) {
...
+		// default address (http://127.0.0.1:4646).
+		addr := n.Address
+		if !strings.Contains(addr, "://") {
+			addr = "http://" + addr
+		}
+		apiCfg.Address = addr
+	}
+
+	return NewNomadSource(n.Priority, apiCfg, n.Path)
+}
```
tgross: Suppose I want to run Pack as a Nomad job itself using the [Task API socket](https://developer.hashicorp.com/nomad/api-docs/task-api). How should we change this to support that?

internal/pkg/variable/source/nomad_source.go
```diff
@@ -0,0 +1,130 @@
...
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/nomad/api"
+
+	"github.com/hashicorp/nomad-pack/sdk/pack"
+	"github.com/hashicorp/nomad-pack/sdk/pack/variables"
+	"github.com/zclconf/go-cty/cty"
+	ctyjson "github.com/zclconf/go-cty/cty/json"
```
tgross: Tidy up your imports

tgross (changes requested):

internal/pkg/variable/source/config.go
DeekshithaTimmareddy: I think there is no need of change in the code and it already supports and works through the env form. So api client already handles sockets. I tested it locally by proxying a unix socket and fetch worked and also i have added comments to make it understand better.

image

image

tgross: >  I tested it locally by proxying a unix socket and fetch worked and also i have added comments to make it understand better.

Ok great, but you left the comment here in the code instead of the usage help text.

(Also, in the future, instead of posting screenshots of terminal text, you can post the text itself. Screenshots of text are hard to read for us old folks :grin:  If you've got a lot of text (as you do here), try wrapping it in a `<details>` block)

DeekshithaTimmareddy: I moved into usage test, Thanks for the tips, it really helping me a lot and i will definitely make a note of this.

tgross (approved):
LGTM

